PR #3, opened 2019-12-17
Fix support 3rd party library completion

heavenshell:
## Problem
Current `rls` settings can't complete 3rd party library.

## Before
Kapture 2019-12-18 at 1 01 53

## After
Kapture 2019-12-18 at 1 05 40


mattn:
Thank you!

PR #5, opened 2019-12-18
feature: Support yaml-language-server and lsp4xml

johejo:
## Note
Installers for Windows are not included.
The yaml-language-server hardly be useless unless schema is configured at workspace_config.

Example config for Kubernetes YAML
```vim
let g:lsp_settings = {
    \ 'yaml-language-server': {
    \    'workspace_config': {
    \       'yaml': {
    \          'completion': v:true,
    \          'hover': v:true,
    \          'validate': v:true,
    \          'schemas': {
    \             'kubernetes': '/*',
    \           },
    \          'format': {
    \             'enable': v:true,
    \           }
    \        },
    \     },
    \  },
    \}
```

I think this setting is worth adding to the document or wiki in the future.


mattn:
Thank you!

PR #7, opened 2019-12-19
Fix can use old vim version

heavenshell:
I know my vim version is too old...
Can use ++once after 8.1.000(I don't know proper version 😉 )

Please review this 🙇


mattn:
Thanks. How about to write the if/else into `autoload/lsp_settings.vim` ?

```
call lsp_settings#register_server_settings(l:settings)
```

autoload/lsp_settings.vim
```diff
@@ -68,3 +68,12 @@ function! lsp_settings#exec_path(cmd) abort
   endif
   return ''
 endfunction
+
+function! lsp_settings#register_server_settings(settings) abort
+  let settings = a:settings
```
heavenshell: If I set `call lsp#register_server(a:settings)`, following error raised.
```
"autoload/lsp_settings.vim" 79L, 2076C
If you want to enable Language Server, please do :LspInstallServer
Error detected while processing User Auto commands for "lsp_setup":
E121: Undefined variable: a:settings
Press ENTER or type command to continue
```
`autocmd` can't take `a:settings` or `l:settings`?

plugin/lsp_settings.vim
```diff
@@ -143,7 +143,11 @@ function! s:vimlsp_setting() abort
     if l:found ==# 0
       exe printf('augroup vimlsp_suggest_%s', l:ft)
         au!
-        exe printf('autocmd FileType %s ++once call s:vimlsp_settings_suggest()', l:ft)
+        if has('patch-8.1.000')
```
heavenshell: Forgot to check version.

heavenshell:
>Thanks. How about to write the if/else into autoload/lsp_settings.vim ?

Thank you for review 👍 
Fixed. PTAL.

autoload/lsp_settings.vim
heavenshell: autocmd can’t take local scope var, so change to script scope var

settings/bash-language-server.vim
```diff
@@ -1,11 +1,12 @@
 augroup vimlsp_settings_bash_language_server
   au!
-  autocmd User lsp_setup ++once call lsp#register_server({
+  let settings = {
```
heavenshell: [ask]
Should I add `l:` prefix?

mattn:
Thanks

PR #8, opened 2019-12-22
Fix solargraph installer

ippachi:
I ran tha `:LspInstallServer` in ruby project. And I got the belaw message.

```
/home/vagrant/.vim/plugged/vim-lsp-settings/installer/install-solargraph.sh: line 10: call: command not found
```

I don't know what is `call` but I think that is not need. So delete it.
And `bundle exec` command should be probably called in where `bundle install` was called. 
So run `cd $DIR` before calling `solraraph stdio`.


mattn:
Thank you

PR #12, opened 2019-12-24
Feature/add servers fortls metals

johejo:
## Description
- fix: imporve npm and pip install. Windows installation is not included...
- proposal: add simple test script, see `test/run.sh`. Sorry, I have no deep knowledge of the protocol. 

## Testing considerations
- Each LanguageServer has slightly different startup commands.
- Consider multiple Language Servers for one language.
- Should test with vim-lsp function (vimscript) instead of simple JSON-RPC for testing. In fact, the editor(LSP Client) sends a little complicated request.


mattn:
Thank you

PR #22, opened 2019-12-25
Fix Rust integration

kuro46:
There were some problems:

- A binary from `cargo install rls` is too old (https://crates.io/crates/rls)  
When executed this, It will exit with no outputs.
Correct way to install is `rustup component add rls rust-analysis rust-src` (from [README](https://github.com/rust-lang/rls#setup) of RLS)
- Path of `rls` detected via `vim-lsp-settings` is incorrect.
When installed Cargo, It creates a binary named `rls` in the `$HOME/.cargo/bin` by default. (And it usually in the `$PATH`)
image
When executed this, Output is below:
image
So `vim-lsp-settings` marks `$HOME/.cargo/bin/rls` as a path of `rust-language-server` but it doesn't work...

Solution:

I removed local installation way for `rls` because it is depended on the toolchain   and other components.


mattn:
Could you please add new section about installation instruction for rls?

kuro46:
OK, I wrote!

mattn:
Thank you

PR #23, opened 2019-12-25
feature: support swift and cobol

johejo:
Swift for linux are available here.
https://swift.org/download


mattn:
Great! Thank you

PR #43, opened 2019-12-29
fix: sh and cmd editing environments set-up.

tsuyoshicho:
.gitattributes and .editorconfig add (sh lf,cmd crlf)


mattn:
Thank you

PR #44, opened 2019-12-29
fortls support windows python

tsuyoshicho:
add windows py check in fortls
(temp fix; currently method is pyls like)


mattn:
Thank you

PR #47, opened 2019-12-29
plugin: save/restore cpo setting

kazukazuinaina:
It is a small patch.

This patch save and restore cpo setting.


mattn:
Thanks your contribution. But recently, I don't set save_cpo for my plugins. If you have something wrong if this plugin does not do save_cpo, please reopen this.

kazukazuinaina:
Ok, I got it. Thanks for checking and comment.

PR #49, opened 2019-12-31
Setup server directory in Vim script

thinca:
All installer script do same things:  Remove the previous server and create a new directory.

This has 2 problems:
1. Reduced maintainability.
2. Changing install directory is too hard.

By this PR, all installer scripts assume invoked at the installing path.

I'll send an other PR to be configurable the installing server path if this PR is accepted.

### DIFF
--- a/installer/install-bash-language-server.cmd
+++ b/installer/install-bash-language-server.cmd
@@ -1,5 +1,3 @@
 @echo off
 
-cd /d %~dp0
-
-call npm_install bash-language-server bash-language-server
+call "%~dp0\npm_install" bash-language-server bash-language-server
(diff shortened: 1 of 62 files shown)


mattn:
Totally LGTM, Please wait my confirmations.

installer/install-bash-language-server.cmd
```diff
@@ -1,5 +1,3 @@
 @echo off
 
-cd /d %~dp0
-
-call npm_install bash-language-server bash-language-server
+call "%~dp0\npm_install" bash-language-server bash-language-server
```
mattn: This was my lazy thing. But could you please add `.cmd` for this `npm_install ?

thinca: I added `.cmd` for all `call "%~dp0\xxx"`.

mattn:
Thank you

PR #51, opened 2019-12-31
fix: Filetype using lower-case

### LINKED ISSUE #50: :LspInstallServer fail for filetype : 'Dockerfile'
## Error:
Dockerfile open, then type `:LspInstallServer` and error occured.

```
function <SNR>61_vimlsp_install_server の処理中にエラーが検出されました:
行    2:
E684: リストのインデックスが範囲外です: 1
```

## Reason:
Currently, filetype check using `settings.json`'s filetype key name.

- Setting                 : dockerfile
- Vim detect and set type : Dockerfile

```
vimlsp_suggest_dockerfile  FileType
    dockerfile
              call s:vimlsp_load_or_suggest( 'dockerfile' )

```

```
set filetype?
  filetype=Dockerfile
```

Vim autocmd run and suggest, but `:LspInstallServer` do not work above reason.

## Solution:
1. json key tune up vim filetype leagal case and capital.
2. filetype ignore case search(usual normalize lowercase? settings and code)

https://github.com/mattn/vim-lsp-settings/blob/796a6b950a3c520ec24139836135c35364cf941d/plugin/lsp_settings.vim#L35-L36

(do not keysetting use, use current filetype directly.)

tsuyoshicho:
Fix #50 : filetype always lower-case work

### DIFF
--- a/plugin/lsp_settings.vim
+++ b/plugin/lsp_settings.vim
@@ -33,7 +33,7 @@ function! s:executable(cmd) abort
 endfunction
 
 function! s:vimlsp_installer() abort
-  let l:ft = split(&filetype, '\.')[0]
+  let l:ft = tolower(split(&filetype, '\.')[0])
   if !has_key(s:settings, l:ft)
     return []
   endif


mattn:
Thank you

PR #52, opened 2019-12-31
Add g:lsp_settings_servers_dir option

thinca:
This option allows you to change the server installation location.

The default value is the same as the existing value.
However, we may reconsider the default value to reduce traps for dein.vim users.
What do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,12 +22,13 @@ You need to install both [vim-lsp](https://github.com/prabirshrestha/vim-lsp) an
 
 ### Notice
 
-If you use plugin manager that is merging  plugins (ex. dein), Please setting stop merging work(ex. dein / merged = 0).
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -22,12 +22,13 @@ You need to install both [vim-lsp](https://github.com/prabirshrestha/vim-lsp) an
 
 ### Notice
 
-If you use plugin manager that is merging  plugins (ex. dein), Please setting stop merging work(ex. dein / merged = 0).
+If you use plugin manager that is merging  plugins (ex. dein), Please setting stop merging work(ex. dein / merged = 0) or set `g:lsp_settings_servers_dir` option.
```
mattn: > set `g:lsp_settings_servers_dir` option to set a different directory from the vim-lsp's default.

thinca: Thanks.
I updated with force-push and remove duplicated `set`.

```diff
-set `g:lsp_settings_servers_dir` option to set a different directory from the vim-lsp's default.
+set `g:lsp_settings_servers_dir` option to a different directory from the vim-lsp's default.
```

mattn:
Thank you.

PR #59, opened 2020-01-03
fix: support multiple npm packages for one language server

johejo:
Supporting multiple npm packages(e.g. typescript-language-server uses typescript-language-server and tsserver).
Currently, the symbolic link of tsserver is broken :cry: .
Windows support is also required.


mattn:
Thank you

PR #62, opened 2020-01-03
Add clojure-lsp root-uri settings(for leiningen, shadow-cljs, deps.edn)

sakuraiyuta:
In some projects, are not always exist `.git/`.
And Clojure/script projects commonly put package-manager config file(leiningen/shadow-cljs/deps.edn) on root path.
So, I added them to `settings/clojure-lsp.vim`.


mattn:
Thank you

PR #63, opened 2020-01-03
feature: support erlang

johejo:
**Caution**
Only some textDocuments are working in my environment.

Other features (see https://github.com/erlang-ls/erlang_ls#features) are not working...


mattn:
Thank you.

PR #67, opened 2020-01-06
fix: package path

fum1h1ro:
https://github.com/OmniSharp/omnisharp-roslyn/releases

OmniSharp package path for OSX is incorrect.
It does not need `-x64`.

Thanks.


mattn:
Thank you

PR #80, opened 2020-01-09
Don't output errors (skip load) on older vim

idbrii:
vim-lsp requires json_encode and so do we. Skip loading if it's not
available.

I did [a similar PR for vim-lsp](https://github.com/prabirshrestha/vim-lsp/pull/271). It looks like timers is not used for vim-lsp-settings so I omitted it.

Tested on:
VIM - Vi IMproved 7.4 (2013 Aug 10, compiled Nov 24 2016 16:43:18)
Included patches: 1-52
Extra patches: 8.0.0056
Modified by pkg-vim-maintainers@lists.alioth.debian.org


mattn:
Thank you

Issue #61, opened 2020-01-03
feature: support project tsserver


johejo:
The tsserver version used by the project often differs each.
The version of typescript used in create-react-app, angular-cli, etc. is different from the version installed by install-typescript-language-server.sh.

Need to be improved to find `root_uri/node_modules/.bin/tsserver`.
https://github.com/mattn/vim-lsp-settings/blob/ac4f61f82fd8d0c4b0cbabd4e364dced0c72c645/settings/typescript-language-server.vim#L5

PR #82, opened 2020-01-10
feature: #61 support project tsserver

johejo:
Before I start this PR, I found such an implementation in typescript-language-server.
https://github.com/theia-ide/typescript-language-server/blob/3114c557b58eb6b286e3c63f68ffe56c26800224/server/src/lsp-server.ts#L75
In other words, it seems that typescript-language-server looks for tsserver in project even without specifying `--tsserver-path`.

But, I tried project that using typesctipt 3.5 with `vim-lsp-settings@master`, The `:LspDocumentDiagnostics` points out optional chain(feature could be use after typesctipt 3.7) error `typescript:Error:1109` .

create an angular project (using typescript 3.5)
```bash
npx -p @angular/cli ng new sample
...
cd sample/
```

create `src/sample.ts` and open it by vim with vim-lsp and vim-lsp-settings.
```ts
let foo: any = {bar: false};
if (foo?.bar) { // This line should be error but no error pointed out.
}
```

I confused but I wanted to support the project tsserver anyway.
So I create this PR.
I don't have in-depth knowledge of both typescript and LSP.
I want someone's reviews and thoughts.

### DIFF
--- a/autoload/lsp_settings.vim
+++ b/autoload/lsp_settings.vim
@@ -80,3 +80,9 @@ function! lsp_settings#root_uri(pattern) abort
   endif
   return lsp#utils#path_to_uri(l:dir)
 endfunction
+
+function! lsp_settings#find_tsserver() abort
+  let l:root_uri = lsp_settings#root_uri(['tsconfig.json'])[7:] " remove file://
+  let l:project_tsserver = l:root_uri . '/node_modules/.bin/tsserver'
+  return executable(l:project_tsserver) ? l:project_tsserver : ''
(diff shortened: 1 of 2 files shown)


wordijp:
I checked about tsserver to use, `--tsserver-path` option was no longer needed, this problem's fix is just a remove it.
This option was given when `typescript` and `typescript-language-server` were installed in separate directories by PRs(#36, #37).
After that, it was installed in the same directory by PR (# 49). This option is no longer needed, but remains.

PR #83, opened 2020-01-11
fix: remove tsserver path option

johejo:
https://github.com/mattn/vim-lsp-settings/pull/82#issuecomment-573280064


mattn:
Thank you.

PR #94, opened 2020-01-21
Add Linux Mint support to clangd installer

pojiro:
Linux Mint is a Linux distribution based on Ubuntu LTS.
The clangd installer support only ubuntu in Linux, this pull request makes it support LinuxMint too.

### DIFF
--- a/installer/install-clangd.sh
+++ b/installer/install-clangd.sh
@@ -13,12 +13,28 @@ case $os in
         ;;
 esac
 
-# Check Ubuntu version
-ubuntu_version=$(lsb_release -a 2>&1 | grep 'Release' | awk '{print $2}')
-
-case $ubuntu_version in
-    14.04|16.04|18.04)
-        platform="linux-gnu-ubuntu-$ubuntu_version"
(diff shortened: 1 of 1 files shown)


thomasfaingnaert:
I think you can search `/etc/upstream-release/lsb-release` for `DISTRIB_RELEASE` as well. That will work across more distributions.

pojiro:
Thanks for your comment.

> That will work across more distributions.

Do you mean other Linux Mint version or other Linux distributions?
If you mean other Linux distributions, I don't know other distributions have `/etc/upstream-release/lsb-release` .

thomasfaingnaert:
I meant other Linux distros. I'm also not sure if those contain `/etc/upstream-release/lsb-release`, so you'll have to check if that file exists.
One thing that checking `/etc/upstream-release/lsb-release` does though, is avoid the need to manually map Linux Mint versions to Ubuntu versions.

In any case, it was just a suggestion. What you have is fine as well.

mattn:
Thank you

PR #100, opened 2020-01-23
Support .svn root markers

idbrii:
Make it easier to add new root markers and support .svn directories as root markers.

I found a typo in the reason root marker, so I threw that in here too.


autoload/lsp_settings.vim
```diff
@@ -1,5 +1,7 @@
 let s:servers_dir = expand('<sfile>:h:h').'/servers'
 
+let lsp_settings#root_markers = ['.git/', '.svn/']
```
mattn: Thanks. Please change to:

```
['.git/', '.svn/', '.hg/', '.bzr/']
```

mattn:
Thank you.

PR #102, opened 2020-01-24
Use the initialization_options value of g:lsp_settings in vim-language-server

sgur:
Currently, there is no way to set `initialization_options` parameter to vim-language-server.

This change use `g:lsp_settings['vim-language-server'].initialization_options` to set `initialization_options` of the language server.


mattn:
Thank you

PR #110, opened 2020-01-29
Add pyls-ms

mattn:


settings/pyls-ms.vim
```diff
@@ -0,0 +1,32 @@
...
+      \ 'root_uri':{server_info->lsp_settings#get('pyls-ms', 'root_uri', lsp_settings#root_uri(g:lsp_settings_root_markers))},
+	  \ 'initialization_options': lsp_settings#get('pyls-ms', 'initialization_options', {
+      \   'analysisUpdates': v:true,
+      \   'asyncStartup': v:true,
+      \   'displayOptions': {},
+      \   'interpreter': {
+      \      'properties': {
+	  \        'InterpreterPath': 'python', 
+      \        'UseDefaultDatabase': v:true,
+      \        'Version': '3.8.1'
```
lithammer: Probably not a great idea to hardcode `InterpreterPath` and `Version`.

mattn: Fixed

settings/pyls-ms.vim
```diff
@@ -0,0 +1,32 @@
...
+      \ 'name': 'pyls-ms',
+      \ 'cmd': {server_info->lsp_settings#get('pyls-ms', 'cmd', [lsp_settings#exec_path('pyls-ms')])},
+      \ 'root_uri':{server_info->lsp_settings#get('pyls-ms', 'root_uri', lsp_settings#root_uri(g:lsp_settings_root_markers))},
+	  \ 'initialization_options': lsp_settings#get('pyls-ms', 'initialization_options', {
+      \   'analysisUpdates': v:true,
+      \   'asyncStartup': v:true,
+      \   'displayOptions': {},
+      \   'interpreter': {
+      \      'properties': {
+	  \        'InterpreterPath': split(system('where python'), "\n")[0],
```
lithammer: No idea what `where` is, but it doesn't (normally) exist on macOS/Linux from what I can tell. Perhaps it's better to use `exepath()`?

```diff
-'InterpreterPath': split(system('where python'), "\n")[0],
+'InterpreterPath': exepath('python'),
```

Unsure how it works together with a virtualenv though.

settings/pyls-ms.vim
```diff
@@ -0,0 +1,32 @@
...
+      \        'UseDefaultDatabase': v:true,
+      \        'Version': matchstr(system('python -V'), '\s\zs\S\+'),
```
lithammer: This will match the trailing newline so you end up with something like `'Version': '3.8.1\n'`.

```diff
-'Version': matchstr(system('python -V'), '\s\zs\S\+'),
+'Version': trim(matchstr(system('python -V'), '\s\zs\S\+')),
```

plugin/lsp_settings.vim
```diff
@@ -169,13 +169,14 @@ function! s:vim_lsp_install_server(ft, command) abort
 endfunction
 
 function! s:vim_lsp_settings_suggest(ft) abort
-  if empty(s:vim_lsp_installer(a:ft))
+  let l:entry = empty(s:vim_lsp_installer(a:ft))
```
itchyny: There's a mistake here. Remove empty() this line.

mattn: Thank you. Fixed.

settings/pyls-ms.vim
mattn: Fixed in master branch

settings/pyls-ms.vim
mattn: ditto

PR #112, opened 2020-01-29
Make g:lsp_settings higher priority than settings.json

xuanduc987:
Config from g:lsp_settings should have higher priority so that user can
enable efm-langserver without edit settings.json directly.


mattn:
Good catch. Thank you

PR #114, opened 2020-01-29
Add go installer/efm-langserver installer

tsuyoshicho:
This PR includes two-item.

* Add go pkg installer
* Add efm-langserver installer

note:efm-langserver installer tried force set filetype `setf json` then `setf _`
Test only in Windows, do not Linux...sorry.


mattn:
Thank you

PR #120, opened 2020-01-29
add fsautocomplete for fsharp

xuanduc987:
Only tested on macOS.
Someone please test on windows and linux


mattn:
Let's go to merge since no one test this. If something wrong, then let's fix it that time. Thank you.

PR #121, opened 2020-01-29
fix language server for _ ft is not started

xuanduc987:
The servers for _ ft were registered too late, because
`s:vim_lsp_load_or_suggest('_')` was called in after `VimEnter`, a.k.a
after vim-lsp started.


mattn:
Thank you

PR #122, opened 2020-01-29
fix: improve go installer / gopls,efm-langserver

tsuyoshicho:
Improve previous PR's setting

* go_install process improve like gopls steps
* efm-langserver requires add (go)
* gopls,efm-langserver depend go_install and new go_install adjust


mattn:
Thank you

tsuyoshicho:
Merge quickly, thanks.

Issue #126, opened 2020-01-30
Installer scripts are not working if bash does not exists on /bin/bash


nyarla:
Hi,

Some Language server installer scripts for unix-like environemt are using hard-coded `/bin/bash` in shebang, but it's not working if bash does not exists `/bin/bash`.

For examples, I'm using [NixOS](https://nixos.org/) Linux and NixOS does not have executable binary on `/bin` or `/usr/bin`, that has executable binary on `/run/current-system/sw/bin`.

I also think that another case, if vim-lsp-settings users does not installed bash, that case is also not working by `/bin/bash` does not exists.

So I think that installer scripts should not use hard-coded `/bin/bash`, I suggest shebang of installer scripts are changes to `/usr/bin/env sh` or `/bin/sh`.

Or another suggestion, it's configurable by vimrc about execute shell for installer scripts.

PR #128, opened 2020-01-30
fix: Fix bash throws bad interpreter error by fix commit about #126

nyarla:
[This commit](https://github.com/mattn/vim-lsp-settings/pull/127/commits/a95b48fd4c275b64a153adf8455714228f733151) is using bash as execute shell for installer scripts, but this fix make new probrem about installer scripts throws `Bad Interpreter` error if `/bin/bash` does not exists.

My pull request is changed shebang to `#!/usr/bin/env bash`, and this code is fixed about `Bad Interpreter` error.

BTW, I think that my pull request does not care of bash is not installed.
So I suggest adding notes about runtime dependences to README?

### DIFF
--- a/autoload/lsp_settings.vim
+++ b/autoload/lsp_settings.vim
@@ -277,8 +277,7 @@ function! s:vim_lsp_install_server(ft, command) abort
     split new
     call termopen(l:entry[1], {'cwd': l:server_install_dir, 'on_exit': function('s:vim_lsp_install_server_post', [l:entry[0]])}) | startinsert
   else
-    let l:args = has('win32') ? ['cmd', '/c', l:entry[1]] : ['bash', '-c', l:entry[1]]
-    let l:bufnr = term_start(l:args, {'cwd': l:server_install_dir})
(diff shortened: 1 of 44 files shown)


mattn:
Probably, replacing `bash -c xxx.sh` to `bash xxx.sh` will fix this issue.

nyarla:
@mattn I tried replaces`bash -c` to `bash` inside vim script, but **it's not fixed some cases** .

For example, `installer/install-vim-language-server.sh` calls `installer/npm_install.sh`, if only replaced `bash -c` to `bash` inside vim script, it does not fixed `installer/npm_install.sh` throws `Bad Interpreter` error.

I think that nearly case about installer scripts calls another scripts, called script probably throws `Bad interpreter` error without changes shebang.

mattn:
Make sense. Okay, will merge this. Could you please add revert of https://github.com/mattn/vim-lsp-settings/commit/9a95bf72eca2c0da2585a7fb3b4361ba230e425f into this pull-request?

nyarla:
Okay, I'm (maybe) done it.

mattn:
Thanks!

PR #133, opened 2020-01-30
Fix: Added Kotlin Language Server to README

hiroyukim:
Since Kotlin is already included in vim-lsp-settings, the documentation was modified.


mattn:
Thank you

PR #143, opened 2020-02-03
Fix default initialization_options of vue-language-server

naoina:
When use vue-language-server on Neovim, I encountered the following errors.

```
Error detected while processing User Autocommands for "lsp_setup":
E121: Undefined variable: v:none
E116: Invalid arguments for function lsp_settings#get
E116: Invalid arguments for function lsp#register_server
E121: Undefined variable: v:none
```

Probably Neovim has not `v:none`, and the configuration schema is different between vim-lsp-settings'one and defined by [vue-language-server's one](https://github.com/vuejs/vetur/blob/master/server/src/config.ts#L3).
This PR fixes the default configuration value to follows to the correct schema, as a result of that, the above errors on Neovim will be fixed.


mattn:
Thank you

PR #145, opened 2020-02-05
Fix list only required plugin

heavenshell:
I think required plugin should be minimum.

Do we need recommend plugin section?


mattn:
Thank you

PR #146, opened 2020-02-06
Add .tsx and .jsx extentions to LspRegisterServer

amadeus:
These filetypes are in use by a few plugins I maintain, and it would be nice for these configs to just work for them.


mattn:
Good catch. Thank you.

PR #153, opened 2020-02-10
Add LspSwitchSourceHeader based on @kent-tri's patch

mattn:
See https://github.com/prabirshrestha/vim-lsp/pull/703

### DIFF
--- a/settings/clangd.vim
+++ b/settings/clangd.vim
@@ -81,5 +81,12 @@ function! s:switch_source_header() abort
   echo 'Switching between source/header ...'
 endfunction
 
-command! LspSwitchSourceHeader call <SID>switch_source_header()
-nnoremap <plug>(lsp-switch-source-header) :<c-u>call <SID>switch_source_header()<cr>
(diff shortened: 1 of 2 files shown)


prabirshrestha:
Header seems specific to c. Should it be called Alternate instead?

Not sure if other language has it.

No strong preference here.

mattn:
@kent-tri What thought about naming?

kent-tri:
@mattn @prabirshrestha I would stick with the name that `clangd` uses as it is specific to that language server as well as being specific to the C family of languages. But, I have no strong preference either.

https://clang.llvm.org/extra/clangd/Extensions.html

kent-tri:
Since this is outside the `vim-lsp` plugin now, should there be a check to confirm that `vim-lsp` is installed before executing the commands that rely on it?

mattn:
Finally,  I renamed it to `:LspDocumentSwitchSourceHeader` since vim-lsp use `Lsp` + `Document` + method names.

PR #154, opened 2020-02-10
Clarify usage in README

### LINKED ISSUE #135: I can get this to work
When I run `:LspInstallServer typescript-language-server`, it responds with `server not found` (I tried different ones: rls, vim, etc).
If I run `:LspInstallServer` without arguments, then it goes and fetches efm-langserver and installs it.
Am I doing something wrong?
Thanks.

augustjanse:
Fixes #135. I had the same problem, and had to search the issues to find the solution.


mattn:
Thank you.

PR #168, opened 2020-02-18
rust-analyzer: add rust-analyzer

h-michael:
ref: https://github.com/rust-analyzer/rust-analyzer/pull/3216

I think it's better to keep ra_lsp_server for a while as a user transition period,
so I add rust-analyzer instead of rename.

### DIFF
--- a/installer/install-rust-analyzer.sh
+++ b/installer/install-rust-analyzer.sh
@@ -1,0 +1,20 @@
+#!/usr/bin/env bash
+
+set -e
+
+os=$(uname -s | tr "[:upper:]" "[:lower:]")
+
+case $os in
+linux)
+  platform="linux"
+  ;;
+darwin)
+  platform="mac"
+  ;;
+esac
+
+version="2020-02-18"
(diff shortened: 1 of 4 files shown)


installer/install-rust-analyzer.sh
```diff
@@ -0,0 +1,20 @@
...
+case $os in
+linux)
+  platform="linux"
+  ;;
+darwin)
+  platform="mac"
+  ;;
+esac
+
+version="2020-02-18"
```
h-michael: rust-analyzer binary have not been released yet.
So this PR is still WIP.

mattn:
Thanks your contribution. Am I better to wait the release rust-analyzer? or merge?

h-michael:
The Install command does not work because the binary has not been released since the binary was renamed.
I watch rust-analyzer so I can respond immediately when released.
What do you think I should do?

mattn:
I'll merge this because someone(you?) probably send pull-requset when the rust_analyzer will be released. Thanks!

PR #185, opened 2020-03-03
Update pyls-ms: 0.5.31

yaegassy:
**osx**

https://pvsc.blob.core.windows.net/python-language-server-stable?restype=container&comp=list&prefix=Python-Language-Server-osx-x64

**win**

https://pvsc.blob.core.windows.net/python-language-server-stable?restype=container&comp=list&prefix=Python-Language-Server-win-x64

**linux**

https://pvsc.blob.core.windows.net/python-language-server-stable?restype=container&comp=list&prefix=Python-Language-Server-linux-x64


mattn:
Thank you

PR #192, opened 2020-03-09
fix: bump terraform-lsp to v0.0.10

johejo:
https://github.com/juliosueiras/terraform-lsp/releases/tag/v0.0.10


mattn:
Thank you

PR #194, opened 2020-03-10
Add golangci-lint-langserver

mattn:


installer/install-golangci-lint-langserver.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+
+set -e
+
+"$(dirname "$0")/go_install.sh" github.com/nametake/golangci-lint-langserver@latest
```
johejo: Prebuilt binaries are avaliable. https://github.com/nametake/golangci-lint-langserver/releases

installer/install-golangci-lint-langserver.sh
```diff
@@ -0,0 +1,6 @@
...
+"$(dirname "$0")/go_install.sh" github.com/golangci/golangci-lint@latest
```
johejo: Same above https://github.com/golangci/golangci-lint#binary

mattn: golangci-lint-langserver does not support Windows binary.

installer/install-golangci-lint-langserver.sh
mattn: Let's merge this and will fix installer when the Windows binary will be released.

PR #196, opened 2020-03-12
fix css-languageserver setting for scss filetype

nao0x2c6:
Hello, thank you for the awesome plugin.
I noticed css-langserver doesn't work when we edit *.scss files.
If set filetype to "sass" in *.scss, the file is treated as the different syntax from the scss syntax.

This pull request just adds scss filetype to some setting files.

### DIFF
--- a/settings.json
+++ b/settings.json
@@ -556,6 +556,14 @@
       ]
     }
   ],
+  "scss": [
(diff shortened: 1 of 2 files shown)


mattn:
Thank you

PR #199, opened 2020-03-18
fix: rename lsp4xml to lemminx

johejo:
lsp4xml was renamed to lemminx.
https://github.com/eclipse/lemminx

Also, upgrade to v0.11.0.


mattn:
Thank you!

PR #201, opened 2020-03-22
Add vscode-css-languageserver-bin to the README

simnalamburt:
vim-lsp-settings supports vscode-css-languageserver-bin out of the box! Let's document this in README.


mattn:
Thank you

PR #206, opened 2020-03-28
Fix dls installer for Windows

karanokuri:
Hello, thank you for the awesome plugin.
I noticed dls installer doesn't work on Windows.

It is because the installation file refer to undefined environment variable.


mattn:
Thank you!

PR #208, opened 2020-03-29
Fix lsp_settings#root_path is not using arguments

Milly:
The arguments of `lsp_settings#root_path` is not used. Fix it.

The function is used in: https://github.com/mattn/vim-lsp-settings/blob/f361e8e88cc6c130885c3ae24991415cfcc649d1/autoload/lsp_settings/profile.vim#L36
However, it always returned project root directory by `g:lsp_settings_root_markers` because this bug.


mattn:
Thank you

PR #210, opened 2020-04-03
Updated Clang: 9.0.0 -> 10.0.0

### LINKED ISSUE #159: Clangd diagnostics can't include stl headers on Mac
I have installed clangd with `:LspInstallServer` but the diagnostic throws errors when I include any standard library files.
`#include <string>` results in:
```
main.cpp|1 col 10| clang:Error:pp_file_not_found:In included file: 'string.h' file not found /Users/maciej/.vim/bundle/vim-lsp-settings/servers/clangd/bin/../include/c++/v1/string.h:60:15: note: error     occurred here 
```
I took a look at the `string` in the `.vim/bundle/vim-lsp-settings/servers/clangd/bin/../include/c++/v1/` and it fails at:
```
#if !defined(_LIBCPP_HAS_NO_PRAGMA_SYSTEM_HEADER)
#pragma GCC system_header
#endif
#include_next <string.h>
```
with the diagnostics:
```
string.h|57 col 13| clang:Warning:-Wpragma-system-header-outside-header:#pragma system_header ignored in main file
string.h|57 col 13| clang:Warning:-Wpragma-system-header-outside-header:#pragma system_header ignored in main file
string.h|60 col 15| clang:Error:pp_including_mainfile_in_preamble:Main file cannot be included recursively when building a preamble
```
I am not sure whether it is my fault, something wrong with the vim-lip-settings or with the llvm code that is downloaded with the `:LspInstallServer`, but the c++ diagnostics doesn't work out of the box.

Also it results in very *superficial* autocompletion of stl, because the clangd can't parse the library and does not complete methods/members of classes.

kvoj:
This resolves #159.

### DIFF
--- a/installer/install-clangd.cmd
+++ b/installer/install-clangd.cmd
@@ -3,14 +3,14 @@
 setlocal
 
 echo Downloading clang and LLVM...
-curl -L -o LLVM-9.0.0-win64.exe "http://releases.llvm.org/9.0.0/LLVM-9.0.0-win64.exe"
+curl -L -o LLVM-10.0.0-win64.exe "https://github.com/llvm/llvm-project/releases/download/llvmorg-10.0.0/LLVM-10.0.0-win64.exe"
 echo Running setup...
 
 REM Run setup as regular user to avoid a UAC popup
 set __COMPAT_LAYER=RUNASINVOKER
 md tmp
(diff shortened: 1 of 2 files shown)


mattn:
Thank you

PR #211, opened 2020-04-04
Fixed bad notations of #210

kvoj:
I realized that I had used bad notations at #210 after the merging.
Sorry for bothering you twice.


mattn:
LGTM, Thanks.

PR #213, opened 2020-04-08
Document g:lsp_settings_root_markers

### LINKED ISSUE #212: Document g:lsp_settings_root_markers
I was working on a Vue.js+TypeScript project managed in a git mono-repo. The vue-language-server always complained about not finding imported modules and consequently couldn't work properly. Apparently, the language server used the top-level directory containing `.git/` as the root, which makes sense most of the time, but in this mono-repo situation, the vue project was actually in a subdirectory, leading to said errors. I was able to solve the issue by setting
```vim
let g:lsp_settings_root_markers = ['.projections.json', '.git', '.git/']
```
where `.projections.json` is a custom root marker, `.git` is for git submodules and `.git/` is for regular git repositories.
After setting this in the vimrc, everything worked as expected.

Unfortunately I couldn't find any documentation of this, only the [relevant lines in the source code](https://github.com/mattn/vim-lsp-settings/blob/5e603c07e6e1ae3eb55dd8e0c13bf278ada237bc/plugin/lsp_settings.vim#L6-L12). It would be great if this was documented somewhere.

cbows:
Closes #212

### DIFF
--- a/README.md
+++ b/README.md
@@ -249,14 +249,8 @@ let g:lsp_settings = {
 \}
 ```
 
-When resolving the root directory for a language servers, this plugin will look
+When resolving the root directory for a language server, this plugin will look
 for directories containing special root markers defined in `g:lsp_settings_root_markers`.
-
-If you need to specify alternative root markers:
-
-```vim
-let g:lsp_settings_root_markers = ['.projections.json', '.git', '.git/']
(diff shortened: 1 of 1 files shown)


mattn:
Cool, Thanks!

PR #214, opened 2020-04-15
fix lemminx installer

johejo:
Latest lemminx jar is available on https://repo.eclipse.org/content/repositories/lemminx-snapshots/org/eclipse/org.eclipse.lemminx/.
But the lemminx snapshot jar on repo.eclipse.org seems to be obsolete when a new version is released.

To avoid confusion, installer refers to lemminx bundled with vscode-xml.
https://github.com/redhat-developer/vscode-xml/releases

The installer for windows is also probably broken, but I'm sorry I don't have a windows environment to debug.

### DIFF
--- a/installer/install-lemminx.sh
+++ b/installer/install-lemminx.sh
@@ -2,17 +2,26 @@
 
 set -e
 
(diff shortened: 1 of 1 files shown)


mattn:
Thank you

PR #216, opened 2020-04-16
Upgrade elixir-ls

### LINKED ISSUE #215: Upgrading elixir-ls
Now elixir-ls move to new repository [here](https://github.com/elixir-lsp/elixir-ls). And now they provide elixir-ls distribution since v0.3.3 (latest version at the moment) to be the same as older repository does.

wingyplus:
Now elixir-ls move to new repository [1]. So this changes change the
download path and bump to a new version (v0.3.3).

[1] https://github.com/elixir-lsp/elixir-ls

Close #215


mattn:
Thank you

PR #219, opened 2020-04-18
Can specify the project root directry

tennashi:
`:LspSettingsLocalEdit` will automatically find the project root if the project is under the control of VCS. However, if my project is not under the control of VCS, I want to specify the project root manually.
For this purpose, I added an argument to `:LspSettingsLocalEdit` to specify the project root.


mattn:
Thank you

PR #220, opened 2020-04-18
Fix the argument of load_schema and remove debug code

tennashi:
Match the argument of load_schema to the server name.


mattn:
Thanks

PR #221, opened 2020-04-22
Add root_uri_patterns for elixir-ls setting

wingyplus:
This changes add mix.exs to elixir-ls root_uri_patterns setting to
detect elixir project for who's working with multi elixir projects
within a single repository.


mattn:
Thank you

Issue #203, opened 2020-03-24
Need vim help docs


dbarnett:
I'd like to be able to type `:help :LspInstallServer`, `:help g:lsp_settings`, etc. in my vim and have it jump to docs about those topics. Currently vim-lsp-settings has no doc/vim-lsp-settings.txt file or other vim helpdoc files.

Milly:
docs added at 6a8686d20bc94445b495048b624cfc0040e397e9

dbarnett:
Thanks! 👍 I noticed some of the links didn't work, so I sent a small PR to update those (#223).

BTW, installation instructions might not make any sense in the local help since it's awkward to even access the help if the plugin isn't already installed. Consider removing those and just having them in the README.

PR #223, opened 2020-04-25
Add docs for :LspSettingsLocalEdit etc, and reformat help

dbarnett:
Follow-up from #203 / 6a8686 to create missing content and tags referenced in the part that says `do |:LspSettingsLocalEdit| or |:LspSettingsGlobalEdit|`, drops "vim-lsp-settings-" prefix on some tags for global commands, and add a table of contents.


mattn:
Thank you

PR #233, opened 2020-05-09
updatePowershellLsp

seike460:
The Zip file "2.0.0-preview.8" was not found on Github.
Path seems to have changed.
I've fixed it to Path to v2.1.0.


mattn:
Thank you

PR #235, opened 2020-05-10
Update dart analysis server installer

tksgi:
Add MacOS installer
and Fix directory path because directory configuration in zip file seems to have changed


mattn:
Thank you

PR #236, opened 2020-05-12
Add support json schema for LspSettingsGlobalEdit

Jagua:
The path of settings.json:

* global: `$HOME/.local/share/vim-lsp-settings/settings.json`
* local: `/path/to/project/.vim-lsp-settings/settings.json`

This PR adds fileMatch pattern for global.


mattn:
Good catch. Thank you.

PR #237, opened 2020-05-16
Detect python version from python-path setting.

Milly:
Call `python -V`, giving preference to the `python-path` setting.


mattn:
Thank you

PR #238, opened 2020-05-16
Use `call` to make `python.bat` callable.

Milly:
If running python via a `bat` file it will silently stop.
Ex. using [pyenv-win](https://github.com/pyenv-win/pyenv-win)


mattn:
Thank you

PR #242, opened 2020-05-21
venv using system-site-packages

Magicwangs:


mattn:
Why do you need this? What OS, version, version of Vim?

mattn:
If you have an issue, please file new issue.

PR #246, opened 2020-05-26
update omnisharp-roslyn to v1.35.2

supermomonga:
Sorry, #244 was a mistake. v1.35.3 doesn't exists yet and v1.35.2 is the right latest version.


mattn:
Thank you

PR #251, opened 2020-06-05
feat: cmake-language-server setting add

tsuyoshicho:
[cmake\-language\-server · PyPI](https://pypi.org/project/cmake-language-server/) minimal install/setting PR

> Alpha Stage, work in progress.

Keep it to beta/stable?


mattn:
Thank you

PR #260, opened 2020-06-19
Fix pip cmd

Milly:
* Use `python -m pip ...` instead `pip ...`.
* Do not use `--user`, it's does not upgraded pip in venv.
* Echo running commands for improve log.

### DIFF
--- a/installer/pip_install.cmd
+++ b/installer/pip_install.cmd
@@ -1,11 +1,16 @@
 @echo off
 
+REM Usage
+REM : pip_install [EXECUTABLE_NAME] [PYPI_NAME]
+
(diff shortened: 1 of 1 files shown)


mattn:
Thanks. I prefer to use lowercase `rem`.

Milly:
> Thanks. I prefer to use lowercase `rem`.

Changes all `REM` to `rem`? (include other than the lines I changed?)

Milly:
*installer\install-clangd.cmd* contains `REM`.

mattn:
Okay, I'll do it my-self. Thank you.

PR #264, opened 2020-06-21
Change clone depth to 1

itchyny:
As title, I changed the clone depth to 1. Note that in `install-erlang-ls.sh` and `install-groovy-language-server.sh` this option is already set.


mattn:
Thank you

PR #265, opened 2020-06-27
Introduce g:lsp_settings_disable_suggestions

ujihisa:
* Problem: When you use lsp for certain filetypes, you see
  "Please do :LspInstallServer to enable Language Server" message
  every time when you open a file that you don't need lsp supports.
  It's hard to disable all the possible filetypes but the one
  you use.
* Solution: Provide a way to opt-out the messages.


mattn:
Thank you

PR #270, opened 2020-06-30
Fix permissions for Jedi language server install

brentyi:
Updated the permissions of `installer/install-jedi-language-server.sh` to match all of the other install scripts -- just needed to be marked as executable.

(I was getting a permission denied error when running `:LspInstallServer jedi-language-server`)


mattn:
Thank you

PR #271, opened 2020-07-04
Fix install-elm-language-server.sh

IzumiSy:
This PR has two changes

1. `LspInstallServer` for Elm succeeds, but vim-lsp fails in starting LSP for Elm, because the name used in linking is different.
2. `elm-language-server` has to run with `--stdio` option in order to work with clients.


mattn:
Thank you

PR #274, opened 2020-07-07
feat: add haskell ide engine(hie) for haskell

colinaaa:
did not provide a installation script since it should be installed from source on most OS which is
quite complicated and takes a lot of time


mattn:
Thank you

PR #276, opened 2020-07-12
Add rust-analyzer to README

masuke5:
I added rust-analyzer to README since it is already included in vim-lsp-settings.


mattn:
Thank you

PR #278, opened 2020-07-14
Add pyls-all language server (installs pyls with its dependencies)

aiotter:
In the current settings, vim-lsp-settings only installs pyls.
The function of pyls is limited unless its dependencies: Rope, Pyflakes, McCabe, pycodestyle, pydocstyle, autopep8, YAPF. (See [Installation](https://github.com/palantir/python-language-server#installation))

pyls searches those dependencies in its `sys.path`. When pyls is installed into a venv (as vim-lsp-settings does), its dependencies must be installed to the same venv if you want to use their functions.
vim-lsp-settings only installs pyls to a venv, not with its dependencies. In this situation, you can jump to the definition and hover the hint, but you can neither lint nor format the code.

This PR enables users to use the functions just after introducing vim-lsp-settings, without locating venv and performing installing pyls dependencies into it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,7 +97,8 @@ Because there is no way to update a server, please run `:LspInstallServer` again
 | JavaScript       | typescript-language-server                             | Yes           |
 | JavaScript       | javascript-typescript-stdio                            | Yes           |
 | JavaScript       | eslint-language-server                                 | Yes           |
-| Python           | pyls                                                   | Yes           |
+| Python           | pyls-all (pyls with dependencies)                      | Yes           |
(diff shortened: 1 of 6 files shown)


heavenshell:
IMO, user should install options by themself.
Some user want use yapf, but other user want use pyls-black etc.

At least I don't want to install `isort` by force.

aiotter:
Added `pyls-all` language server, which installs pyls with its dependencies. This will be a default LS for filetype `python` when executing `:LspInstallServer`.

I couldn't test it on Windows because I don't have Windows environment. Can anyone help me with testing it on Windows?

mattn:
thank you

PR #282, opened 2020-07-16
add Julia to the supported language list in README

### LINKED ISSUE #279: Add julia to Supported Languages list in README
I think the supported languages list in README should be updated because julia is already supported in #157

AtsushiSakai:
To fix #279 
I added Julia to the supported language list in README


mattn:
Thank you

PR #284, opened 2020-07-18
Update vls config

hrsh7th:
This PR aims to avoid error when using vetur.

For example, `vetur.format.options` is required.

BTW, currently vim-lsp does not support `dynamicRegistration` feature so this PR does not enable format on vetur with vim-lsp now.


mattn:
thank you

PR #286, opened 2020-07-22
Update clojure lsp version to latest

mracos:
Maybe we could also use the [`LATEST` alias](https://github.com/snoe/clojure-lsp/releases/latest) but that would require a little bit more of investigation on how to do it since just replacing `release-$version` with it does not work 😞


mattn:
Thank you

PR #291, opened 2020-07-30
Add angular and more in README

yaegassy:
- Add angular-language-server.
- Sorting the order of "Supported Languages".
- Add missing html-langserver and sqls in "Supported Languages".


mattn:
Thank you!

PR #296, opened 2020-08-07
Add `ocaml-lsp`

yasuo-ozu:
It adds [`ocaml-lsp`](https://github.com/ocaml/ocaml-lsp) (recommended cli name is `ocamllsp`), which is official and brand new language server implementation.

I'm sorry that the PR doesn't include installer for Windows, so future contributions are hoped. (But it works fine for POSIX sys.)

Because simple binary packages are not provided officially, it compiles binaries on local. It will not pollute user's local file system, but it requires `opam` command and `git` command.
(`curl | tar`, alternative to `git`, will not work here because GitHub's zip-download feature will drop some of `ocaml-lsp`'s submodules.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,6 +123,7 @@ Because there is no way to update a server, please run `:LspInstallServer` again
 | Lua              | sumneko-lua-language-server                            | Yes           |
 | Nim              | nimls                                                  | No            |
 | PHP              | intelephense                                           | Yes           |
+| OCaml            | ocaml-lsp                                              | Yes           |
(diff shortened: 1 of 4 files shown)


mattn:
Thank you. Could you please fix some shellcheck errors?

https://github.com/mattn/vim-lsp-settings/pull/296/files

yasuo-ozu:
@mattn Thanks for your review. Fixed it.

mattn:
Thank you

PR #297, opened 2020-08-11
Add rome

heavenshell:
Maybe rome can use lint and format.

Install
Kapture 2020-08-12 at 1 23 00

Lint and Format
Kapture 2020-08-12 at 1 21 11


mattn:
Thank you

PR #320, opened 2020-10-11
Update Metals versions and initializationOptions

ckipp01:
This pr just updates the default version of Metals to the latest. It also adds in the `isHttpEnabled` initializationOption and sets it to true. This allows the [Metals Doctor](https://scalameta.org/metals/docs/editors/new-editor.html#metals-http-client) to work in clients that don't have an embedded html viewer.


mattn:
Thank you

PR #323, opened 2020-10-15
Install solargraph without development gems

chiastolite:
```
$ cd ~/.local/share/vim-lsp-settings/servers/solargraph

# before
$ bundle list --name-only |wc
      39      39     397

# after
$ bundle list --name-only |wc
      22      22     219
 ```


mattn:
Thank you

PR #324, opened 2020-10-15
Update README.md

fgr1986:
`pyls-all` is the default server name after installation, so configurations should refer to it instead of `pyls`.


mattn:
Thanks!

PR #329, opened 2020-10-28
Fixed a bug in cl-lsp and enabled

cxxxr:
See https://github.com/cxxxr/cl-lsp/commit/a14d5b93576d4fb820a1ffdd7497198e85eb37fb

### DIFF
--- a/settings.json
+++ b/settings.json
@@ -489,7 +489,6 @@
   "lisp": [
     {
       "command": "cl-lsp",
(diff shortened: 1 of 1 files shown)


mattn:
Sorry. I noticed this pull-request just now. As you know, currently I'm having an issue about ros or cl-lsp. It seems that the issue occur only for me. So I reverted the change.  🙇

PR #330, opened 2020-10-29
feat: global setting file store customizable

tsuyoshicho:
## Global config customize support

This PR is below change:

- remove lsp_settings#data_dir() (only using by lsp_settings#profile#edit_global())
- add    lsp_settings#global_settings_dir()
- fix global setting file direct operation change to using lsp_settings#global_settings_dir()

and provide global setting dir config : `g:lsp_settings_global_settings_dir'

### pros
- inner module(profile) not by used data_dir(low layer setting) directly, use just provided function's return dir path
- customizable global setting path

### cons
- I think that nothing.

### DIFF
--- a/autoload/lsp_settings.vim
+++ b/autoload/lsp_settings.vim
@@ -17,16 +17,20 @@ call remove(s:settings, '$schema')
 
 let s:ftmap = {}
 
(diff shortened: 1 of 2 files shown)


mattn:
Thank you

PR #342, opened 2020-11-18
Add ntt (TTCN-3 language server)

5nord:
Add ntt, a language server for TTCN-3.

### DIFF
--- a/installer/install-ntt.sh
+++ b/installer/install-ntt.sh
@@ -4,7 +4,6 @@ set -e
 set -o pipefail
 
 version="0.4"
-filename="ntt.tgz"
 os="$(uname -s | tr "[:upper:]" "[:lower:]")"
 
 case "${os,,}" in
--- a/installer/install-ntt.sh
+++ b/installer/install-ntt.sh
@@ -1,0 +1,22 @@
+#!/bin/bash
+
+set -e
+set -o pipefail
+
+version="0.4"
+filename="ntt.tgz"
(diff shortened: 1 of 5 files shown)


installer/install-ntt.sh
```diff
@@ -0,0 +1,22 @@
...
+
+set -e
+set -o pipefail
+
+version="0.4"
+os="$(uname -s | tr "[:upper:]" "[:lower:]")"
+
+case "${os}" in
+darwin|linux)
+  os="$(echo "$os" | sed -e "s/\b\(.\)/\u\1/g")"
```
mattn, line 11: remove this line.

installer/install-ntt.sh
```diff
@@ -0,0 +1,22 @@
...
+  url="https://github.com/nokia/ntt/releases/download/v${version}/ntt_${os}_x86_64.tar.gz"
```
mattn, line 12: use `"${os^}"`

5nord: This shell expansion is not supported on Darwin platforms, because they only provide bash versions 3.x

installer/install-ntt.sh
5nord: I'll change the download scheme, so we can get rid of this unfortunate line.

mattn:
Thank you.

5nord:
I thank you! Your project is very helpful to me.

PR #346, opened 2020-11-26
Handle dependency error in a better way

subnut:
If server is found, but dependencies are not found, then currently
"Server not found" is shown, making user clueless. This commit tries to
inform the user that server was found, but dependencies were not
satisfied.


mattn:
Thank you

Issue #348, opened 2020-11-30
Suggestions Not Showing


thecontinium:
I am using Neovim on Mac. 
I install this with both` brew install neovim` or `brew install neovim --HEAD` and get the same results for both.

My init.vim just loads two plugins:

```
call plug#begin('~/.config/nvim/plugged')
Plug 'prabirshrestha/vim-lsp'
Plug 'mattn/vim-lsp-settings'
call plug#end()
```

I start neovim with: `nvim -V9/tmp/messages` and then in vim:
`:e foo.clj`
I am expecting the message `Please do :LspInstallServer to enable Language Server clojure-lsp` to show but there is **no output** and `:messages` shows **nothing**

In `/tmp/messages` there is an entry:
```
autocommand call s:vim_lsp_load_or_suggest('clojure')
Please do :LspInstallServer to enable Language Server clojure-lsp
```

Some help would be appreciated. Cheers.

mattn:
strange.

> In /tmp/messages there is an entry:

This mean that you can shows the messages with :message.

thecontinium:
That's what I thought. 

Also if I put some debug message in the first line of the function `vim_lsp_load_or_suggest` => `echom "ft=" . a:ft ` I only get the first call in the initialisation with `_` output in the messages `ft=_`

No other calls are recorded when I change file type.
If I change the echom to a call to input I get the input message every time !!

Any ideas ?

thecontinium:
This looks to be the reason why ...

https://github.com/neovim/neovim/wiki/FAQ#calling-inputlist-echomsg-etc-in-filetype-plugins-and-autocmd-does-not-work

thecontinium:
fixed

PR #349, opened 2020-12-02
unsilent echom called from au FileType

thecontinium:
See: https://github.com/mattn/vim-lsp-settings/issues/348


mattn:
Thank you

mattn:
Thank you

PR #351, opened 2020-12-13
Update cl-lsp installer

cxxxr:
The old version bundled lem, but it was too hard to manage, so I decided to install them separately.


mattn:
Thank you

PR #357, opened 2020-12-24
Support V programming Language

zakuro9715:
Support [V Programming Lanugage](https://vlang.io)

[vlang/vls](https://github.com/vlang/vls) is very experimental but it have some useful features even now.

I have no windows knowledge. I try write windows installer, but I don't know if it will work correctly. Please verify it.

Unix installer works for me. But because v and vls is experimental and v depends on c compiler, build may fail in some environments.

### DIFF
--- a/installer/install-vlang-vls.sh
+++ b/installer/install-vlang-vls.sh
@@ -1,0 +1,7 @@
+#!/usr/bin/env bash
+git clone --depth=1 https://github.com/vlang/vls .
+
+echo 'Compiling vlang/vls...'
+v -prod cmd/vls
+mv cmd/vls/vls vlang-vls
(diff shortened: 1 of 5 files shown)


installer/install-vlang-vls.sh
```diff
@@ -0,0 +1,6 @@
+#/usr/bin/env bash
+git clone --depth=1 https://github.com/vlang/vls .
```
mattn, line 2: missing `!`

mattn (changes requested):

installer/install-vlang-vls.sh
zakuro9715: Oops! 
Fixed :pray:

mattn:
LGTM, Thank you.

PR #359, opened 2020-12-25
Add haskell-language-server

odzhychko:
Even though the haskell-language-server project provides usable static binaries, I didn't create installers. That is because different versions of binaries must be downloaded, depending on the version of the used GHC (Glasgow Haskell Compiler).


mattn:
Thank you

PR #374, opened 2021-01-10
update texlab to v2.2.1

opuyoshi:


opuyoshi:
This version seems unstable.

PR #375, opened 2021-01-14
Updated Omnisharp to 1.37.5

lilwayne1556:


mattn:
I reverted change for catalog.json. It should be a copy of content from schemastore.

https://raw.githubusercontent.com/SchemaStore/schemastore/master/src/api/json/catalog.json

Anyway, thanks.

PR #385, opened 2021-02-01
Update clojure-lsp repo and version

mracos:


installer/install-clojure-lsp.cmd
```diff
@@ -1,5 +1,5 @@
 @echo off
 
 setlocal
-set VERSION=20200706T152722
```
mracos, line 4: I'm not that well versed with windows' cmd, so I'm not sure if this is right 😄

mattn: Thanks. But please remove double-quotes.

mracos: [done here](https://github.com/mattn/vim-lsp-settings/pull/385/commits/de33e9544979f4a707915964bdc09cd87fc67b5d)!

mattn:
Thank you!

PR #393, opened 2021-02-16
Add support for http(s) proxy when pulling metals artifacts with java

matawed:
Behind a corporate proxy I couldn't install metals lsp due to proxy issues.

This PR adds support for proxy when calling java binary in installer/install-metals.sh


mattn:
Thank you. I'll update install-metals.cmd for Windows in later.

PR #398, opened 2021-02-18
Update ntt release URL

5nord:
This PR replaces the old TTCN-3 language server URL with one pointing always to the latest language server version.


mattn:
Thank you

PR #402, opened 2021-02-26
Add deno goto definition and cache file install

heavenshell:
https://user-images.githubusercontent.com/56591/109324012-78fa5a80-7897-11eb-979c-a299a51f5b77.mp4

## Summary

- Add GoTo definition command for deno lsp.
- Add install all cache file command

## Detail

Currently vim-lsp can't goto cached file such as deno's std library, third party library.
https://github.com/denoland/deno/tree/main/cli/lsp#custom-requests

We should request virtual text document to deno lsp.

## Help wanted

I don't have any Windows env, it would be very appricated to check/review Windows point of view 🙏


mattn:
Thank you!

PR #410, opened 2021-04-01
Update terraform-ls version to latest ( 0.15.0 )

netmarkjp:
0.15.0 has released at 2021-03-12
https://github.com/hashicorp/terraform-ls/releases

In my local PC, 0.15.0 works well.
(Please tell me if there is another requirement to merge. :pray: )


mattn:
Thank you

PR #412, opened 2021-04-14
chore: bump tailwindcss-intellisense to 0.5.10

MindTooth:
This fixes using [Just-in-Time Mode](https://tailwindcss.com/docs/just-in-time-mode) in TailwindCSS 2.1.

As I bumped the `.cmd` file on macOS, I had this strange symbol when diffing the change.  Is that something that I can solve myself?

Screen Shot 2021-04-14 at 18 25 34

I'm running 0.5.10 just fine using the download script.  Using vanilla `vim-lsp` though.

### DIFF
--- a/installer/install-tailwindcss-intellisense.cmd
+++ b/installer/install-tailwindcss-intellisense.cmd
@@ -1,7 +1,7 @@
 @echo off
 
 setlocal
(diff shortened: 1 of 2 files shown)


mattn:
Thank you

PR #414, opened 2021-04-20
Add esbonio

heavenshell:
https://user-images.githubusercontent.com/56591/115434530-c4036f00-a243-11eb-82a9-dc6dc9aa3d26.mp4

LSP of Sphinx

Ref 
- https://github.com/swyddfa/esbonio
- https://github.com/yaegassy/coc-esbonio

TODO
- [x] Add windows installer
- [x] Fix README

Help wanted
I don't have Windows and not familiar with cmd.
It would be very appreciated to try or review .cmd file 🙏

### DIFF
--- a/installer/install-esbonio.sh
+++ b/installer/install-esbonio.sh
@@ -1,0 +1,17 @@
+#!/usr/bin/env bash
+
+set -e
+
+python3 -m venv ./venv
+./venv/bin/pip3 install -U pip
+./venv/bin/pip3 install "esbonio[lsp]"
+./venv/bin/pip3 install "pygls==0.9.1"
+
+cat <<EOF >esbonio
+#!/usr/bin/env bash
+
+DIR=$(cd $(dirname $0); pwd)
+$DIR/venv/bin/python3 -m esbonio
+EOF
+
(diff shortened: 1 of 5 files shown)


installer/install-esbonio.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
+
+set -e
+
+python3 -m venv ./venv
+./venv/bin/pip3 install -U pip
+./venv/bin/pip3 install "esbonio[lsp]"
+./venv/bin/pip3 install "pygls==0.9.1"
```
yaegassy, line 8: It is no longer necessary to install a fixed version of pygls. 😇 <https://github.com/swyddfa/esbonio/issues/147>

heavenshell: Thank you! I'll remove this

mattn:
Thanks!

ejoerns:
Was there a specific reason to put `python` in `requires` instead of `python3`?

It refuses to install on my Debian since `python` is not available there. Other settings also use `python3`, thus it should be more consistent and probably work in more environments?

PR #415, opened 2021-04-28
Fix typo

pojiro:
Currently this function  is unused, I think author intended that  the name is "installed_servers".


mattn:
Thank you

PR #416, opened 2021-04-28
Improve local var name

pojiro:
s:settings[l:ft] return list. So I think it is better to name the variable "servers" rather than "server".


mattn:
Thank you

PR #417, opened 2021-04-28
Use latest release for omnisharp

lilwayne1556:
Changed to use the latest release for omnisharp. I haven't ran into any issues with updating manually to the latest version when they're released


mattn:
Thank you

PR #420, opened 2021-05-03
feat: vim-language-server latest config apply

tsuyoshicho:
# Description

## Detail

Add below "initializationOptions" params:

- isNeovim : check has('nvim')
- iskeyword : copy form `iskeyword` option (isk) that appended ":"
- diagnostic.enable : set true

see https://github.com/iamcco/vim-language-server

And add config "refresh_pattern"

--

### Pros

- neovim work fine
- completion work "s", "s:", "s:f".... suggeste to "s:funcname".
- and also work all other g:, l:....

### Cons

Nothing for my investigated result.

## Tuning

If reviewer think needing to add/remove config, please comment.

### DIFF
--- a/settings.json
+++ b/settings.json
@@ -1090,6 +1090,9 @@
       "requires": [
         "npm"
       ],
+      "config": {
+        "refresh_pattern": "\\([bwtglsav]:\\)\\?\\k*$"
+      },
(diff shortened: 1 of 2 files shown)


mattn:
Thank you

Issue #311, opened 2020-09-15
Install specific version of a server


mathieux51:
Howdy folks,

Is it possible to install a specific version of a server? It could look like this:

```
:LspInstallServer terraform-lsp@0.0.11-beta2
```

where `v0.0.11-beta2` is a Github release.

For now I do it manually and it works but it would be convinient to be able to do it in vim.

musl:
I would also love to be able to do this and switch between versions. I have some projects that need Java 11 and some that need Java 8.

PR #425, opened 2021-05-25
[POC] Support installing the server with custom version

mracos:
### Why?

There are some LSP servers that have a steady update rate, and in order to not flood this repository with PR just to update the version I through that having custom-defined versions to install would be nice

### How it's done?

First we need to add the configuration as a `version` key in the language server configuration, for example:

```vim
let g:lsp_settings = { 'clojure-lsp': { 'version': '2021.05.22-16.50.45' } }
```

Then this version is retrieved in the language-server install process and is passed to the installer script as an argument when called, e.g. `installer/install-clojure-lsp.sh 2021.05.22-16.50.45`

Not sure if this the best approach ☝️  but is the one with the smaller footprint that I could think of

I already changed the `clojure-lsp` installer to test this approach and has worked nicely.

### Caveats

Each installer script (that install a specific version) would need to be changed in order to support this custom version argument, however, it would not break for current not changed installer scripts since the argument would be ignored

### Next steps?

Not sure, maybe change all installer scripts to support this argument? Another different approach? Being ok with having to send PRs to this repository 

Update: kinda related to #311?

### DIFF
--- a/installer/install-clojure-lsp.cmd
+++ b/installer/install-clojure-lsp.cmd
@@ -1,5 +1,5 @@
 @echo off
 
 setlocal
-set VERSION=2021.02.01-20.37.52
+set VERSION=%1:-2021.02.01-20.37.52%
 curl -L -o clojure-lsp.cmd https://github.com/clojure-lsp/clojure-lsp/releases/download/%VERSION%/clojure-lsp
--- a/autoload/lsp_settings.vim
+++ b/autoload/lsp_settings.vim
@@ -113,10 +113,14 @@ function! s:vim_lsp_installer(ft, ...) abort
     if l:missing !=# 0
       continue
     endif
+
     if lsp_settings#executable(l:command)
-      return [l:conf.command, l:command]
+      let l:version = lsp_settings#get(l:conf.command, 'version', '')
+      return [l:conf.command, l:command . ' ' . l:version]
     endif
+
   endfor
+
   return [v:false] " placeholder, so that empty() returns false, but len() < 2 returns true
 endfunction
 
--- a/installer/install-clojure-lsp.sh
+++ b/installer/install-clojure-lsp.sh
@@ -2,6 +2,6 @@
 
 set -e
 
-version="2021.02.01-20.37.52"
+version=${1:-"2021.02.01-20.37.52"}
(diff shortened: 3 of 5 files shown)


installer/install-clojure-lsp.cmd
```diff
@@ -1,5 +1,5 @@
 @echo off
 
 setlocal
-set VERSION=2021.02.01-20.37.52
+set VERSION=%1:-2021.02.01-20.37.52%
```
mracos, line 4: I have no idea if this works, I used this [tool](https://daniel-sc.github.io/bash-shell-to-bat-converter/) to convert a bash expression to a bat one haha

mattn:
Seems good for me.

mracos:
Great! I'll in the next following days:

1. Add support for the `version` argument to language server installer scripts that install a specific version (and check if it could work for ones that install directly from `main`)
2. Add some documentation and examples around this new option that can be passed to language servers in `g:lsp_settings`

Maybe I'm missing something? Do you see anything else that could be done?

mattn:
> Maybe I'm missing something? Do you see anything else that could be done?

I think this is good. But I wonder some installers always install latest version. What should do in this PR for them.

mracos:
> I think this is good. But I wonder some installers always install latest version. What should do in this PR for them.

The only case where I see that we would have some work is installer scripts that _do not support_ a specific version ([solargraph](https://github.com/mattn/vim-lsp-settings/blob/master/installer/install-solargraph.sh) for example), we could change them to install from channels where we can pass a specific version but default to lastest.

In other cases where the script already supports but only chooses to pass the latest (e.g. scripts that download from github releases), we can just pass the specific version defaulting to lastest.

WDYT, does it make sense?

mattn:
I would like to gather the opinions of other's for this. Anyone, thought?

mracos:
Maybe an alternative would be to always try and install the `latest` lsp server instead of having to support custom versions.

For example, my use case used to be that the version for `clojure-lsp` pinned here was behind at least by a couple of versions, where the newer ones had fixes that made a huge difference.

If we always install the `latest` for all plugins, I could just run `LspInstallServer` and everything would be rainbows.

WDYT?

mattn:
Closing this old POC — it now has conflicts and the approach was never settled. Thanks for the idea; a fresh take on version pinning is still welcome.

PR #426, opened 2021-05-26
Update elixir-ls version to 0.7.0

chsanch:
The used version of [elixir-ls](https://github.com/elixir-lsp/elixir-ls) 0.6.2 does not work with the latest version of Elixir, it's better to use the latest version (0.7.0)


mattn:
Thank you!

PR #430, opened 2021-05-30
fix: npm (un)install with whitespace paths

djdv:
When doing 
`:LspInstallServer yaml-language-server`
vim-lsp-settings is trying to install the server to: `C:\Users\Dominic Della Valle\AppData\Local\vim-lsp-settings\servers\yaml-language-server`.

But fails with:
```
Installing yaml-language-server
Error detected while processing function <SNR>63_vim_lsp_install_server:
line   28:
CreateProcess failed
Press ENTER or type command to continue
```
because the command being executed is interpreted as `C:\Users\Dominic` due to the space in the username.

This fixes it, but I'm not familiar with Vimscript.
If there's a more conventional way to join the paths for `mkdir` or translate them before passing them to `termopen`|`term_start` it can be changed. But otherwise I think it will be okay.

The changes to the `.cmd` file are similar to https://github.com/mattn/vim-lsp-settings/pull/319 
They do the same thing as the changes to the `.vim` file, but in batch instead of Vimscript.

### DIFF
--- a/autoload/lsp_settings.vim
+++ b/autoload/lsp_settings.vim
@@ -188,10 +188,11 @@ function! lsp_settings#exec_path(cmd) abort
   if type(l:paths) == type([])
     let l:paths = join(l:paths, ',') . ','
   endif
-  let l:paths .= lsp_settings#servers_dir() . '/' . a:cmd
   if !has('win32')
+    let l:paths .= lsp_settings#servers_dir() . '/' . a:cmd
     return lsp_settings#utils#first_one(globpath(l:paths, a:cmd, 1))
   endif
+  let l:paths .= lsp_settings#servers_dir() . '\' . a:cmd
   for l:ext in ['.exe', '.cmd', '.bat']
     let l:path = globpath(l:paths, a:cmd . l:ext, 1)
     if !empty(l:path)
@@ -298,6 +299,9 @@ function! s:vim_lsp_uninstall_server(command) abort
   endif
   call lsp_settings#utils#msg('Uninstalling ' . a:command)
   let l:server_install_dir = lsp_settings#servers_dir() . '/' . a:command
+  if has('win32')
+     let l:server_install_dir = substitute(l:server_install_dir, '/', '\', 'g')
+  endif
   if !isdirectory(l:server_install_dir)
     call lsp_settings#utils#error('Server not found')
     return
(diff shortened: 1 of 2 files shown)


mattn:
Thanks your contribution. For the changes for autoload/lsp_settings.vim, could you please try to call `shellescape(l:entry[1])` ?

```diff
diff --git a/autoload/lsp_settings.vim b/autoload/lsp_settings.vim
index 121f26e..aa654cc 100644
--- a/autoload/lsp_settings.vim
+++ b/autoload/lsp_settings.vim
@@ -353,9 +353,9 @@ function! s:vim_lsp_install_server(ft, command, bang) abort
   call lsp_settings#utils#msg('Installing ' . l:entry[0])
   if has('nvim')
     split new
-    call termopen(l:entry[1], {'cwd': l:server_install_dir, 'on_exit': function('s:vim_lsp_install_server_post', [l:entry[0]])}) | startinsert
+    call termopen(shellescape(l:entry[1]), {'cwd': l:server_install_dir, 'on_exit': function('s:vim_lsp_install_server_post', [l:entry[0]])}) | startinsert
   else
-    let l:bufnr = term_start(l:entry[1], {'cwd': l:server_install_dir})
+    let l:bufnr = term_start(shellescape(l:entry[1]), {'cwd': l:server_install_dir})
     let l:job = term_getjob(l:bufnr)
     if l:job != v:null
       call job_setoptions(l:job, {'exit_cb': function('s:vim_lsp_install_server_post', [l:entry[0]])})
```

djdv:
@mattn 
Thanks for the quick reply.

I made the change and it seems to work in vim, and nvim on my machine.

djdv:
(Sorry quick whitespace edit)

mattn:
Thank you. One more question. `fnamemodify` is required?

djdv:
Good look. It seems like they are not required.
Originally I saw that the path being returned from `lsp_settings#exec_path` looked like this: `C:\Users\Dominic Della Valle\AppData\Local\vim-lsp-settings\servers/yaml-language-server`
(notice the last slash)

But Windows does not care about the mixed slash, so we can do without it.

mattn:
> But Windows does not care about the mixed slash, so we can do without it.

Ah, this is that I want to know.  I'm sorry to bother you. Could you please fix for fnamemodify again?

djdv:
No problem. 
It seems like fnamemodify needed to be modified anyway.
It should look closer to: `fnamemodify('C:\hello/world', ':p:s?\/?\?g')`

The previous patch was not actually converting the slashes. It was just making sure the path is "full".
I'll double check those lines and their output now.

djdv:
Okay.
exec_path, install, and uninstall, should all return more native filepaths now.

I avoided using `s` in `fnamemodify` in favor of calling `substitute` separately.
Other parts of the code already do that, and it would require an `else` to declare the variable on non win32.
Plus the line becomes longer.

mattn:
Thank you

PR #442, opened 2021-07-08
Don't consider servers which are disabled or excluded for suggestion or for install

merrilymeredith:
When opening a perl file with all default settings, despite efm-langserver being disabled by default, it is still shown as an install suggestion.  Similarly, if you run `:LspInstallServer` it offers to install efm-langserver.

Perl currently has 3 servers with no installers defined, plus efm-langserver added to the end of the list, and all are checked regardless of the disabled status or being excluded by a user-defined `g:lsp_settings_filetype_...` variable, so efm-langserver is checked and suggested for install.

This case isn't as noticeable for filetypes with no servers defined or with installers available, but does show up for languages where there are installable servers that have been disabled by user preference.

This PR duplicates the checks that happen in `vim_lsp_load_or_suggest` to the similar loop in `vim_lsp_installer` which is searching for the first available installer.  The second commit adds a warning when the user requests a specific server to be installed and that server is currently disabled by config - they _could_ have installed it before this change but they'll want to fix their config in order to use that server anyway.

### DIFF
--- a/autoload/lsp_settings.vim
+++ b/autoload/lsp_settings.vim
@@ -91,11 +91,30 @@ function! s:vim_lsp_installer(ft, ...) abort
     let l:servers += s:settings['_']
   endif
 
+  let l:default = get(g:, 'lsp_settings_filetype_' . a:ft, '')
+
   let l:name = get(a:000, 0, '')
   for l:conf in l:servers
     if !empty(l:name) && l:conf.command != l:name
       continue
     endif
+
+    if lsp_settings#get(l:conf.command, 'disabled', get(l:conf, 'disabled', 0))
+      continue
+    endif
+
+    if type(l:default) ==# v:t_list
+      if len(l:default) ># 0 && index(l:default, l:server.command) == -1
+        continue
+      endif
+    elseif type(l:default) ==# v:t_string
+      if !empty(l:default) && l:default != l:server.command
+        continue
+      endif
+    else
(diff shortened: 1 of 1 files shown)


autoload/lsp_settings.vim
```diff
@@ -91,11 +91,42 @@ function! s:vim_lsp_installer(ft, ...) abort
...
 
+  let l:default = get(g:, 'lsp_settings_filetype_' . a:ft, '')
+
   let l:name = get(a:000, 0, '')
   for l:conf in l:servers
     if !empty(l:name) && l:conf.command != l:name
       continue
     endif
+
+    if lsp_settings#get(l:conf.command, 'disabled', get(l:conf, 'disabled', 0))
```
mattn, line 102: I don't like global default value `get(l:conf, 'disabled', 0)` since it should always be command name.

merrilymeredith: I am not sure about this line, but it mirrors [autoload/lsp_settings.vim#L463](https://github.com/mattn/vim-lsp-settings/blob/52ae22b7800b132bb841fbfbed525297617df162/autoload/lsp_settings.vim#L463) right now.  Perhaps both need to be fixed?

mattn: Ah, sorry. This is my fault.

mattn:
Thank you

PR #444, opened 2021-07-18
Fix the permission

rnakamine:
When I tried to use ruby_language_server, I didn't have the permission to run the installer, so I fixed the permission.


heavenshell (approved):
LGTM!

mattn:
Thank you

PR #447, opened 2021-08-13
Fix apex-jorje-lsp installer.

shinonome128:
This command is redirected and as a result the installation fails.

command
`curl -o apex-jorje-lsp.jar "https://github.com/forcedotcom/salesforcedx-vscode/blob/develop/packages/salesforcedx-vscode-apex/out/apex-jorje-lsp.jar?raw=true"
`

response
`<html><body>You are being <a href="https://github.com/forcedotcom/salesforcedx-vscode/raw/develop/packages/salesforcedx-vscode-apex/out/apex-jorje-lsp.jar">redirected</a>.</body></html>
`

Adding the curl'-L' option will get the redirected content and the installation will succeed.


mattn:
Thank you

PR #455, opened 2021-09-02
add Volar to Support Languages table on README

tsukkee:
Sorry but I forgot Volar description to the table on #454


mattn:
Thank you

PR #458, opened 2021-09-03
update Volar to 0.27.14

tsukkee:
Update Volar from 0.27.12-alpha.1 to 0.27.14.

From 0.27.14, volar exposes command so the installers are simplified by using npm_install.sh/cmd.


mattn:
Thank you

PR #470, opened 2021-09-27
Add new commands for taplo-lsp

kkiyama117:
# About
Add two commands
- `LspTomlToJson` gets the whole text of the current buffer and generates json.
- `LspJsonToToml` parse current buffer and generates toml

# Todo
- Support parsing text from range selection or given arguments.
- Make `LspJsonToToml` available if not taplo is loaded.


mattn:
Looks nice.

mattn:
Thank you

PR #473, opened 2021-09-28
Add installer file of taplo

kkiyama117:
I took the challenge to add the installer of taplo-lsp.  I'm not sure of windows' one working well but probably works on Linux(at least in my environment).


kkiyama117:
I checked Windows' one probably works on real machine but need to be checked by anyone since I'm not sure of it working.

mattn:
Thank you

PR #475, opened 2021-09-30
To Fix "Error in contrib.url(repos, type)"

ysaito8015:
Set download url to get avoid the error `Error in contrib.url(repos, type) : trying to use CRAN without setting a mirror`


mattn:
Thank you

PR #477, opened 2021-10-03
Revert "fix to use yarn in yaml-language-server"

itchyny:
This reverts commit 25b05bcd98a9c13e99e72f8ee5e8350d271e40b1.

Since yaml-language-server is available via npm, we can use npm to
install the server again. And npm command is more available than yarn.

Ref: https://github.com/redhat-developer/yaml-language-server/issues/521
Ref: https://github.com/redhat-developer/yaml-language-server/pull/528


mattn:
Thank you

PR #481, opened 2021-10-09
Update installing terraform-ls fixed version to latest version

dakatan:
Currently, terraform-ls is installed with a fixed version, but I changed it to install the latest version.
If this is acceptable, remove terraform-ls and run `:LspInstallServer` again, and it will be up-to-date, making it easier to upgrade.


mattn:
Than you. Please check some linter outputs.

PR #491, opened 2021-11-09
Fix deno

mattn:
related #490 
image


hrsh7th:
this has a side-effect that will causes the mismatch when the completing `obj.property|`, I think.

mattn:
Thank your pointing this.

PR #493, opened 2021-11-11
Update arch name

jinpei0908:
Thank you for supporting aarch64 architecture. However `uname -m` output for apple aarch64 may be `arm64`.

Here is my m1 MacBook Air output.

```bash
$ uname -ms
Darwin arm64
```

And also shown below.

>Ok, apparently the output is like this:
Darwin MacBook-Pro.local 20.1.0 Darwin Kernel Version 20.1.0: Sat Oct 31 00:07:10 PDT 2020; root:xnu-7195.50.7~2/RELEASE_ARM64_T8101 arm64
https://developer.apple.com/forums/thread/668206

### DIFF
--- a/installer/install-rust-analyzer.sh
+++ b/installer/install-rust-analyzer.sh
@@ -19,7 +19,7 @@ linux)
 darwin)
   if [[ $arch == "x86_64" ]]; then
     platform="x86_64-apple-darwin"
-  elif [[ $arch == "aarch64" ]]; then
+  elif [[ $arch == "arm64" ]]; then
     platform="aarch64-apple-darwin"
   else
     echo "unknown architecture: $arch"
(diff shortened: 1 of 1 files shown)


installer/install-rust-analyzer.sh
```diff
@@ -19,7 +19,7 @@ linux)
 darwin)
   if [[ $arch == "x86_64" ]]; then
     platform="x86_64-apple-darwin"
-  elif [[ $arch == "aarch64" ]]; then
+  elif [[ $arch == "arm64" ]]; then
```
mattn, line 22:
```
  elif [[ $arch == "aarch64" -o $arch == "amd64" ]]; then
```
I'm not well for apple mac but is this better? or not?

jinpei0908: Thank you for your suggestion. I think it's better. I will update.

mattn:
Thank you

PR #498, opened 2021-11-23
Support deno.json

heavenshell:
Since deno v1.14, deno can set deno.json or deno.jsonc.
If deno.json found at root project, change blocklist.


mattn:
Thanks!

PR #499, opened 2021-11-30
Support deno import intellisense

heavenshell:
https://github.com/prabirshrestha/vim-lsp/pull/1236

This update can use deno import intellisense.

https://user-images.githubusercontent.com/56591/144055474-3b521d5d-8e3e-4981-860b-90d0c4c02e7a.mp4

And also support`reloadImportRegistries` to reload cached response.

### DIFF
--- a/settings/deno.vim
+++ b/settings/deno.vim
@@ -16,6 +16,18 @@ augroup vim_lsp_settings_deno
       \     'test': v:true,
       \     'testArgs': ['--allow-all'],
       \   },
+      \   "suggest": {
+      \     "autoImports": v:true,
+      \     "completeFunctionCalls": v:true,
+      \     "names": v:true,
+      \     "paths": v:true,
+      \     "imports": {
+      \       "autoDiscover": v:false,
+      \       "hosts": {
+      \         "https://deno.land/": v:true,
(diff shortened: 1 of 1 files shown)


settings/deno.vim
```diff
@@ -16,6 +16,18 @@ augroup vim_lsp_settings_deno
       \     'test': v:true,
       \     'testArgs': ['--allow-all'],
       \   },
+      \   "suggest": {
+      \     "autoImports": v:true,
+      \     "completeFunctionCalls": v:true,
+      \     "names": v:true,
+      \     "paths": v:true,
+      \     "imports": {
+      \       "autoDiscover": v:false,
```
heavenshell, line 25: [memo]
https://github.com/denoland/vscode_deno/blob/main/docs/ImportCompletions.md#configuration
This setting is off at vim-lsp-settings.
Because if value is true, editor shoud promped like `add registry url?` and if select `yes`, editor should add registry url to `hosts` setting dynamic.
vim-lsp-setting does not provide such function.

mattn: make sense

mattn:
Thank you.

PR #501, opened 2021-12-04
Add g:markdown_fenced_languages to deno section

heavenshell:
Add interaction to README for avoid LspHover show error message deno lsp

ref: https://github.com/hrsh7th/vim-vital-vs/pull/14


mattn:
Good catch.

mattn:
Thank you.

PR #502, opened 2021-12-05
Update elixir-ls version to 0.9.0

pojiro:
ref. [Release v0.9.0](https://github.com/elixir-lsp/elixir-ls/releases/tag/v0.9.0)


mattn:
Thank you

PR #504, opened 2021-12-12
Update README.md

ryicoh:
To use Deno, We need to set g:lsp_settings_filetype_typescript.

https://vim-jp.slack.com/archives/C01N4L5362D/p1639293510416700


mattn:
Thank you

PR #506, opened 2021-12-15
Fix `lsp_settings#merge`

ichizok:
`lsp_settings#merge` doesn't merge settings.

e.g. yaml-language-server settings:

https://github.com/mattn/vim-lsp-settings/blob/cb25a3685f6c1cee08ff01c6add8f78a1fd5fc46/settings/yaml-language-server.vim#L11

in this case, `lsp_settings#merge(...)` calls and returns `extend(g:lsp_settings['yaml-language-server'], {'yaml': ...})`
but this processing overwrites user settings `g:lsp_settings['yaml-language-server']` by the default values,
so, as a result, user settings is ignored.

This PR replaces `extend` in `lsp_settings#merge` by `s:extend` at lsp_settings/profile.vim, which can merges settings as expected.

### DIFF
--- a/autoload/lsp_settings.vim
+++ b/autoload/lsp_settings.vim
@@ -167,7 +167,7 @@ function! lsp_settings#merge(name, key, default) abort
   if type(a:default) ==# v:t_func
     return extend(l:config, a:default(a:name, a:key))
   endif
-  return extend(l:config, a:default)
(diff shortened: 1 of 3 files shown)


mattn:
Ah, right. I missed that case. Thanks.

PR #507, opened 2021-12-16
Fix vim-language-server asyncomplete refresh pattern

hrsh7th:
Currently, the completion will be invoked when the cursor is onto the empty line.


mattn:
Thank you

PR #518, opened 2021-12-25
Add support for open-browser.vim

4513ECHO:
I made the plugin that it use [tyru/open-browser.vim](https://github.com/tyru/open-browser.vim) in `:LspManageServers` if exists.


mattn:
Thank you

PR #522, opened 2021-12-25
[improve] update example auto-complete plugin

hashue:
## What I did ?
change example plugin from deoplete.nvim to ddc.vim.
## Why
Because author of deoplete.nvim said 
>Note: The development of this plugin is finished. Accepts minor patches and issues but no new features. ddc.vim is the next generation auto completion plugin. Consider migrating to it.


mattn:
Thank you

PR #531, opened 2022-01-19
Update install-eclipse-jdt-ls.sh

maka-io:
Update to use HTTPS


mattn:
Could you please fix some warnings in PR?

https://github.com/mattn/vim-lsp-settings/pull/531/files

mattn:
Thank you

PR #532, opened 2022-01-25
Fix warnings about the deprecated options of bundle

ichizok:
Use `bundle config set --local` to set `path` and `without` flags.


mattn:
I also wanted to fix it. thank you.

PR #535, opened 2022-02-09
modify rust installer for MINGW64

GutsTang:
Modify install-rust-analyzer.sh to support MINGW64.

### DIFF
--- a/installer/install-rust-analyzer.sh
+++ b/installer/install-rust-analyzer.sh
@@ -26,6 +26,20 @@ darwin)
     exit 1
   fi
   ;;
+mingw64_nt*)
+  if [[ $arch == "x86_64" ]]; then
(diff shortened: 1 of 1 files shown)


mattn:
Do you want to install Windows rust-analyzer on non-Windows OSs?

GutsTang:
No. 
My comment is not clear.
I installed MSYS2 in my windows PC. It's my main develop environment.
When I edit my rust code in MSYS2 via VIM. It will think I'm working in Linux.
The script `install-rust-analyzer.sh` is called instead of `install-rust-analyzer.cmd`.
However `uname` will return `MINGW64_NT-***` in MSYS2.

Therefore, `install-rust-analyzer.sh` cannot get the correct platform to install the LSP Server.

mattn:
Thank you

PR #536, opened 2022-02-09
Fix fsautocomplete installer excutable fetching

gihyeonsung:
[fsautocomplete](https://github.com/fsharp/FsAutoComplete) stopped releasing server executables directly for now.

Since 

- a maintainer [recommends installing the server via package manager](https://github.com/fsharp/FsAutoComplete/issues/883) and
- installer is unavailable for now,

I propose to update the installer to get the server via `dotnet tool install` (nuget package manager).

I updated and checked the installer shell script on macos but couldn't do this for batchfile because I have no windows machine. Sorry for this.

ref: fsharp/FsAutoComplete/issues/883

### DIFF
--- a/installer/install-fsautocomplete.sh
+++ b/installer/install-fsautocomplete.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+set -e
+
 if command -v dotnet >/dev/null 2>&1; then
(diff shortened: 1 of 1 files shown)


mattn:
Thank you

PR #539, opened 2022-02-19
Fix eslint-language-server repository in `settings.json`

Qs-F:
This PR fixes the minor problem and does NOT affect the code's behavior.

The problem is, https://www.npmjs.com/package/eslint-language-server is different from https://github.com/microsoft/vscode-eslint .
Although the source code for `eslint-language-server` is seemed to be retrieved from https://github.com/microsoft/vscode-eslint , `settings.json` indicates the wrong repository.

It looks like this URL does not impact the behavior of the tool.
However, while I inspected the `eslint-language-server`, mistakenly saw the wrong repository because of the `settings.json`. So decided to contribute.

Lastly, thanks for your fantastic work!

### DIFF
--- a/settings.json
+++ b/settings.json
@@ -599,7 +599,7 @@
     },
     {
       "command": "eslint-language-server",
-      "url": "https://www.npmjs.com/package/eslint-language-server",
+      "url": "https://github.com/microsoft/vscode-eslint",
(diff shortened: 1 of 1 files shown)


mattn:
Thank you

Issue #540, opened 2022-02-21
awk: syntax error installing terraform-ls on MacOS


bobdoah:
MacOS comes with a version of awk which doesn't support the third, array, argument to the match function. This is a GNU extension.

The install fails with:
```
awk: syntax error at source line 1
 context is
        match($0, /"tag_name": >>>  "(.+)"/, <<<
awk: bailing out at source line 1
~
```

[The change to introduce arch compatability](https://github.com/mattn/vim-lsp-settings/commit/386d493ff17513d57b11fb5ccca4363bb7fa9b44) introduced this issue.

mattn:
@halostatue can you handle this issue?

bobdoah:
@mattn I've put together a PR which undoes the incompatible part of @halostatue's change.

halostatue:
PR #541 works as expected. I’ll need to suggest the same thing for tfvm.

mattn:
Closed by #541

PR #541, opened 2022-02-21
Use compatible awk expression

bobdoah:
Undo the awk expression change introduced in PR #537 . The third argument
is not supported by the non-GNU awk installed on MacOS by default.

Fixes issue #540

### DIFF
--- a/installer/install-terraform-ls.sh
+++ b/installer/install-terraform-ls.sh
@@ -25,7 +25,7 @@ arm64*) arch=arm64 ;;
 esac
 
 version=$(basename $(curl -Ls -o /dev/null -w %\{url_effective\} https://github.com/hashicorp/terraform-ls/releases/latest))
-short_version=$(echo "$version" | awk '{print substr($0, 2)}')
+short_version=$(echo "$version" | cut -c2-)
 filename="terraform-ls_${short_version}"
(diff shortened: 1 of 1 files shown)


installer/install-terraform-ls.sh
```diff
@@ -24,11 +24,8 @@ arm64*) arch=arm64 ;;
   ;;
 esac
 
-version=$(
-  curl --location --silent "https://api.github.com/repos/hashicorp/terraform-ls/releases/latest" |
-    awk 'match($0, /"tag_name": "(.+)"/, a) { print a[1] }'
-)
-short_version=$(echo "${version}" | awk 'match($0, /v(.+)/, a) { print a[1] }')
+version=$(basename $(curl -Ls -o /dev/null -w %\{url_effective\} https://github.com/hashicorp/terraform-ls/releases/latest))
+short_version=$(echo "$version" | awk '{print substr($0, 2)}')
```
mattn, line 28: How about `echo $version | cut -c2-` ?

bobdoah: That works fine. I only swapped it back to what was there before, as I thought that would be preferred.

halostatue:
This works. Apologies for the incompatibility.

mattn:
Thank you

Issue #527, opened 2022-01-09
Cannot find `remark-parse` in remark-language-server


s417-lama:
I tried remark-language-server added by in @remcohaszing in #516, but I got the following error.

```
Could not find module remark-parse
Error [ERR_MODULE_NOT_FOUND]: Cannot find package 'remark-parse' imported from /path/to/workingdir
    at node_internal_ (file://[$HOME]/.local/share/vim-lsp-settings/servers/remark-language-server/node_modules/import-meta-resolve/lib/errors.js:315:11)
    at __node_internal_ (file://[$HOME]/.local/share/vim-lsp-settings/servers/remark-language-server/node_modules/import-meta-resolve/lib/errors.js:258:13)
    at new NodeError (file://[$HOME]/.local/share/vim-lsp-settings/servers/remark-language-server/node_modules/import-meta-resolve/lib/errors.js:242:5)
    at packageResolve (file://[$HOME]/.local/share/vim-lsp-settings/servers/remark-language-server/node_modules/import-meta-resolve/lib/resolve.js:968:9)
    at moduleResolve (file://[$HOME]/.local/share/vim-lsp-settings/servers/remark-language-server/node_modules/import-meta-resolve/lib/resolve.js:1024:18)
    at defaultResolve (file://[$HOME]/.local/share/vim-lsp-settings/servers/remark-language-server/node_modules/import-meta-resolve/lib/resolve.js:1062:13)
    at resolve (file://[$HOME]/.local/share/vim-lsp-settings/servers/remark-language-server/node_modules/import-meta-resolve/index.js:27:12)
    at attempt (file://[$HOME]/.local/share/vim-lsp-settings/servers/remark-language-server/node_modules/load-plugin/index.js:179:15)
    at resolvePlugin (file://[$HOME]/.local/share/vim-lsp-settings/servers/remark-language-server/node_modules/load-plugin/index.js:160:16)
    at addModule (file://[$HOME]/.local/share/vim-lsp-settings/servers/remark-language-server/node_modules/unified-engine/lib/configuration.js:338:18)
    at use (file://[$HOME]/.local/share/vim-lsp-settings/servers/remark-language-server/node_modules/unified-engine/lib/configuration.js:319:13)
    at addEach (file://[$HOME]/.local/share/vim-lsp-settings/servers/remark-language-server/node_modules/unified-engine/lib/configuration.js:288:53)
[...]

PR #545, opened 2022-03-02
Prevent `remark-language-server` for `markdown`

### LINKED ISSUE #543: Restrict or remove `remark-language-server`
After extensive discussion, I believe that it is a _mistake_ that `remark-language-server` has been added to `vim-lsp-settings` as a "markdown" language server. The maintainers there do not seem to be interested in supporting the use case of "people who use markdown", but instead are exclusively interested in supporting the use case of "people who use remark".

See: https://github.com/remarkjs/remark/discussions/960

I believe that the correct answer for this is to _either_ prevent `remark-language-server` from being suggested as a language server for markdown (instead only doing so if both markdown _and_ remark configuration minimums have been matched, e.g., https://github.com/neovim/nvim-lspconfig/pull/1641). If that particular complexity cannot be met, then I recommend that `remark-language-server` be _removed_ from `vim-lsp-settings` because it is not fit for the purpose of being a language server for markdown without opting into the entire `remark` ecosystem.

It should only be reconsidered for expansion if they change their stance, which I have no reason to believe will happen based on the discussion noted above. As it is, the suggested installation of `remark-language-server` is a misleading option _at best_ with the current settings.

halostatue:
Resolves #543, #527

The current configuration for `remark-language-server` has it proposed for installation whenever a markdown file is edited. This is incorrect, as `remark-language-server` is _only_ for use in a `remark` project, which implies a lot of extra baggage (node-base infrastructure, configuration, etc.) not expected by most users.

As it does appears that `vim-lsp` and `vim-lsp-settings` allow only filetype server identification by default, this PR:

- removes the `markdown` entry from the `allowlist` for   `remark-language-server`;
- removes the `markdown` entry from `settings.json` (as `remark-language-server` was the only LS there, and
  `remark-language-server` is _not_ a markdown language server);
- and updates the compatibility listing so that `remark-language-server` is noted as `Markdown (remark)`.

My personal preference would be to wholly _remove_ `remark-language-server` (it provides value to a very small group of
developers), but as it _may_ be possible to extend the `allow` test in the future so as to detect appropriate configuration, there is currently no reason to remove it, so long as it is possible to install `remark-language-server` where required.

Current testing suggests that this is _not_ possible, and a modification to `settings.json` to put `node_modules/.bin/remark` 
in the `requires` block for `remark-language-server` won’t work as it is not necessary to install `remark-cli` in a remark project. (That is, it _does_ appear to work, but it does _not_ appear to work _well_, and we still see the stupid error message from `remark-language-server`.)

If we want to keep `remark-language-server` (I personally see negative value in it) for any use whatsoever, then I believe that `vim-lsp-settings` _and/or_ `vim-lsp` need to be modified so that it will only be suggested for installation or started if preconditions are met.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,99 +79,99 @@ You can change the directory to install servers by set `g:lsp_settings_servers_d
 
 ## Supported Languages
 
-| Language         | Language Server                     | Installer | Local Install |
-|------------------|-------------------------------------|:---------:|:-------------:|
-| Apex/VisualForce | apex-jorje-lsp                      | Yes       | Yes           |
-| Bash             | bash-language-server                | Yes       | Yes           |
-| C#               | omnisharp                           | Yes       | Yes           |
-| C/C++            | clangd                              | Yes       | Yes           |
-| COBOL            | cobol-language-support              | Yes       | Yes           |
-| CSS              | css-languageserver                  | Yes       | Yes           |
-| CSS              | tailwindcss-intellisense            | Yes       | Yes           |
-| Clojure          | clojure-lsp                         | Yes       | Yes           |
-| Clojure          | clj-kondo-lsp                       | Yes       | Yes           |
-| Cmake            | cmake-language-server               | Yes       | Yes           |
-| D                | dls                                 | Yes       | No            |
-| D                | serve-d                             | Yes       | No            |
-| Dart             | analysis-server-dart-snapshot       | Yes       | Yes           |
-| Dockerfile       | dockerfile-language-server-nodejs   | Yes       | Yes           |
-| Dot              | dot-language-server                 | Yes       | Yes           |
-| Elixir           | elixir-ls                           | Yes       | Yes           |
-| Elm              | elm-language-server                 | Yes       | Yes           |
-| Erlang           | erlang-ls                           | Yes       | Yes           |
-| F#               | fsharp-language-server              | Yes       | Yes           |
-| Fortran          | fortls                              | Yes       | Yes           |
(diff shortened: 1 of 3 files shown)


mattn:
Looks good to me. Ready to merge?

halostatue:
Yes, I think so. I believe that enhancements to `vim-lsp` or `vim-lsp-settings` should be considered to allow it to be reenabled as I indicated in my last paragraph:

> If we want to keep remark-language-server … for any use whatsoever, then I believe that vim-lsp-settings and/or vim-lsp need to be modified so that it will only be suggested for installation or started if preconditions are met.

This change makes `remark-language-server` unusable _by default_ (one cannot do `LspInstallServer remark-language-server`). I _believe_ that individuals who use `remark-language-server` with `vim-lsp` can add a local vimrc configuration where it can be reenabled, but as I am not one of those individuals, I am not well placed to suggest how it might happen.

@remcohaszing may have some suggestions on this until context-enabling enhancements can be made to allow `remark-language-server` to be used only in `remarkjs` projects. I would also be unopposed to _restoring_ `remark-language-server` if and when it can be used as a general-purpose markdown language server.

remcohaszing:
I’m working upstream to add some options so `remark-language-server` so it can no-op for projects that haven’t configured remark. For the time being I agree that it might be best to temporarily disable it.

I don’t think the language in the table should be changed to `Markdown (remark)`. remark and `remark-language-server` work on regular markdown files. This notation isn’t used for other language servers that integrate a tool either (i.e. eslint or flow).

mattn:
Thanks.

PR #549, opened 2022-03-16
Fix build error for ocaml-lsp

ushitora-anqou:
For OCaml (*.ml) files, the command `:LspInstallServer` fails to build ocaml-lsp. This is because how to build ocaml-lsp has changed recently, and we need to install its dependencies by `make install-test-deps` rather than `opam switch create . --with-test` (see also https://github.com/ocaml/ocaml-lsp/pull/638). Also, the built executable is located at a different place now. This PR fixes these problems.


mattn:
Thank you

PR #550, opened 2022-03-18
Fix refresh pattern to be customizable

yoshi1123:
In `settings.json`, eclipse-jdt-ls has this config:
```
      "config": {
        "refresh_pattern": "\\([.a-zA-Z0-9_-]\\+\\|/\\|\\k\\+\\)$"
      },

```
Completion for methods/fields does not work for me with this refresh pattern.
With this PR, one can change this refresh_pattern with `g:lsp_settings`. For example,
```
let g:lsp_settings = {
\ "eclipse-jdt-ls": {
\   "config": {
\     "refresh_pattern":"\\(\\k\\+\\)$"
\   }
\ }
\}
```

### DIFF
--- a/autoload/lsp_settings.vim
+++ b/autoload/lsp_settings.vim
@@ -472,6 +472,11 @@ function! s:vim_lsp_load_or_suggest(ft) abort
   if get(g:, 'lsp_loaded', 0)
     for l:server in s:settings[a:ft]
       let l:config = lsp_settings#server_config(l:server.command)
+      if has_key(g:, 'lsp_settings') &&
+      \  has_key(g:lsp_settings, l:server.command) &&
+      \  has_key(g:lsp_settings[l:server.command], 'config')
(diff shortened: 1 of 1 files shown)


autoload/lsp_settings.vim
```diff
@@ -472,6 +472,11 @@ function! s:vim_lsp_load_or_suggest(ft) abort
   if get(g:, 'lsp_loaded', 0)
     for l:server in s:settings[a:ft]
       let l:config = lsp_settings#server_config(l:server.command)
+      if has_key(g:, 'lsp_settings') &&
+      \  has_key(g:lsp_settings, l:server.command) &&
+      \  has_key(g:lsp_settings[l:server.command], 'config')
+        call extend(l:config, g:lsp_settings[l:server.command]['config'])
+      endif
```
mattn, line 479: extend is destructive API. please override `refresh_pattern` simply. (not to use extend)

PR #551, opened 2022-03-22
Update volar 0.33.6

tsukkee:
update volar to v0.33.6

* add rename step to the installer scripts because the original command name was changed
  * see https://github.com/johnsoncodehk/volar/blob/master/CHANGELOG.md#0330
* set volar to the first option for .vue file since Volar is the official recommendation in current Vue community 
  * see https://blog.vuejs.org/posts/vue-3-as-the-new-default.html

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,8 +165,8 @@ You can change the directory to install servers by set `g:lsp_settings_servers_d
 | TypeScript        | eslint-language-server              |    Yes    |      Yes      |
 | Vim               | vim-language-server                 |    Yes    |      Yes      |
 | Vala              | vala-language-server                |    No     |      No       |
-| Vue               | vue-language-server                 |    Yes    |      Yes      |
 | Vue               | volar-server                        |    Yes    |      Yes      |
+| Vue               | vue-language-server                 |    Yes    |      Yes      |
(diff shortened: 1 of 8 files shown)


tsukkee:
@mattn I have some questions about this PR.

* From volar 0.33.0, volar's command name was changed to `vue-language-server` (see https://blog.vuejs.org/posts/vue-3-as-the-new-default.html for more details). In my understanding, vim-lsp-settings has command name based naming convention, so I changed file names related to volar as this draft PR. But, of course, this can be breaking change for users (probably I am only one user to use this though... 😅). Is this right way to handle this situation?

* Now Vue community is going to Vue 3 ( https://blog.vuejs.org/posts/vue-3-as-the-new-default.html ), and official recommended language server is also being changed from Vetur (vls) to Volar. So, I want to change default LSP for .vue file from vls to Volar in vim-lsp-settings by updating `settings.json`. Does this look good?

mattn:
1. Please rename them. We can't take care of changing project name. Thanks.
2. Yes, please do.

tsukkee:
Thank you!

> 1. Please rename them. We can't take care of changing project name. Thanks.
> 2. Yes, please do.

I've done these.

mattn:
Thank you

PR #552, opened 2022-03-26
Remove `-range` from `nnoremap <plug>(lsp-rust-join-lines)`

cuviper:
AFAIK, `-range` is not a valid option on maps. This caused me to have a
broken mapping for `-range` itself, adding an annoying timeout on my
unrelated `-` mapping.

```
:nmap -
n  -range      * <Plug>(lsp-rust-join-lines) :<C-U>call <SNR>131_join_lines()<CR>
n  -             <Plug>(dirvish_up)
```


cuviper:
Maybe it should have `<buffer>` like all the other maps here though?
(and `command! -buffer` as well)

mattn:
@prabirshrestha What your thinking ?

mattn:
Thank you

PR #562, opened 2022-04-28
Update install-metals.sh to always install the latest release

yuilib:
Changed the shell script to always download a proper release version of metals automatically. But as for install-metals.cmd, I don't have any knowledge about .cmd, so it wasn't changed.
Also because the previous commit 895408f7a022e9226ce6b0387d958ac20a52ae36 was wrong, the version specified in the .cmd script is reverted to 0.11.2.


mattn:
Thank you.

PR #571, opened 2022-05-12
Update volar to v0.34.11

tsukkee:
Update volar version and update its setting according to the latest volar's change.


mattn:
Thank you

PR #575, opened 2022-05-22
Force yaml formatting to be enabled

asmka:
Relates: https://github.com/redhat-developer/yaml-language-server/issues/486

yaml-language-server can use prettier as a formatter.
This PR forces formatting capability to be enabled, so `:LspDocumentFormat` works in yaml files.


mattn:
Thank you

PR #587, opened 2022-06-21
Fix hie and hle configs

a5ob7r:
haskell-ide-engline(hie) and haskell-language-server(hls) are usually called via their wrapper executables. However this plugin looks up non-wrapper executables but not wrapper them in loading configurations, so it doesn't load configs for hie and hls even if wrappers are installed.

To resolve this, change command name which is used in loading configs from non-wrappers to wrappers, and add configs for wrappers instead of not-wrappers. And for compatibility, keep up loading user's non-wrapper configs.


a5ob7r:
@mattn I think that this CI error isn't related to this patch. Could you rerun CI jobs?

https://github.com/mattn/vim-lsp-settings/runs/6978143577?check_suite_focus=true

```
Error: Could not download and unarchive asset https://github.com/vim/vim-win32-installer/releases/download/v8.2.5141/gvim_8.2.5141_x64.zip at C:\Users\RUNNER~1\AppData\Local\Temp\vim-installer: request to https://github.com/vim/vim-win32-installer/releases/download/v8.2.5141/gvim_8.2.5141_x64.zip failed, reason: socket hang up
Error: Could not download and unarchive asset https://github.com/vim/vim-win32-installer/releases/download/v8.2.5141/gvim_8.2.5141_x64.zip at C:\Users\RUNNER~1\AppData\Local\Temp\vim-installer: request to https://github.com/vim/vim-win32-installer/releases/download/v8.2.5141/gvim_8.2.5141_x64.zip failed, reason: socket hang up
```

mattn:
Yes, it is another issue. Thanks.

a5ob7r:
Thank you, too.

PR #589, opened 2022-06-24
Add support for texlab aarch64-apple-darwin

yoshida-ryuhei:
By this pull request https://github.com/latex-lsp/texlab/pull/591, TexLab began to distribute the binary for aarch64-apple-darwin (unfortunately not for Linux aarch64).
If you could merge this patch, aarch64-apple-darwin or M1 and M2 mac users can download TexLab using vim-lsp-settings.

I would appreciate it if you could check this pull request.


mattn:
Thank you

PR #595, opened 2022-07-10
installers bash -> POSIX /bin/sh

krzysckh:
this PR
- changes `#!/usr/bin/env bash` || `#!/bin/bash` to `#!/bin/sh` for installer shebangs
- makes installer scripts POSIX compatible (by replacing bashisms)

```sh
$ shellcheck *.sh
```
returns nothing, so all the scripts should work as before

why?

i wanted to install `go` tools on my openbsd laptop without bash ;)

### DIFF
--- a/installer/go_install.sh
+++ b/installer/go_install.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/bin/sh
 
 # Usage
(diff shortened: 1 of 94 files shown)


mattn:
Thank you for this PR. I'll look this in later.

mattn:
Thank you

PR #597, opened 2022-07-15
Fix; Removed redundant line that caused exit 127

varuld:
The variable 'ubuntu_version' does on ubuntu 18.04 caused error full exit with exit code
of 127, the removal of this fixes the error


mattn:
Thank you

Issue #601, opened 2022-07-26
Installer install-fsharp-language-server.sh is broken


ibsenrune:
When installing the `fsharp-language-server`, you run into a couple of issues:

* The install script `install-fsharp-language-server.sh` does not specify what folder to build the language server to. However, the launch script (generated by the install script) assumes that the application is being built with .NET Core 2 putting it in a netcoreapp2.0 folder. This fails at the time of writing because .NET 6 is now the newest version and the application will be in a net6 folder.

* On a brand new machine with .NET recently installed, the `paket` tool will not be available, so building the [fsprojects/fsharp-language-server](https://github.com/fsprojects/fsharp-language-server) source code will fail.

* When attempting to build the sharp-language-server source code, the install script will build the whole solution, including unit tests etc. We might speed up the process and make it more robust by only building what we need. As an example, at the time of writing, there is a non-compiling test project in [fsprojects/fsharp-language-server](https://github.com/fsprojects/fsharp-language-server), which means that the install will fail for no good reason when vim-slp-settings installs the language server.

* The launch script attempts to invoke the assembly `FSharpLanguageServer.dll`. But that file is not an executable and must be launched by `dotnet`.

PR #602, opened 2022-07-26
Fix install-fsharp-language-server.sh, cf. issue #601

ibsenrune:
The purpose of this PR is to fix the issues described in #601:

* We make the install and launch scripts agnostic to the .NET version by publishing the built application to a Publish folder (in other words, the target folder name is not dependent on the .NET version number). 
* We make sure the `paket` tool is installed and available.
* We only build the `FSharpLanguageServer` project, ignoring any other projects like test projects.
* We launch the built application using dotnet `FSharpLanguageServer.dll`


mattn:
Thank you.

Issue #566, opened 2022-05-06
Install Scala Language Server failed.


wangzhihao:
`:LspInstallServer` against `*.scala`

```
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100 42577  100 42577    0     0  72042      0 --:--:-- --:--:-- --:--:-- 72042
Resolution error: Error downloading org.scalameta:metals_2.12:0.11.5
  not found: /home/zhihaow/.ivy2/local/org.scalameta/metals_2.12/0.11.5/ivys/ivy.xml
  not found: https://repo1.maven.org/maven2/org/scalameta/metals_2.12/0.11.5/metals_2.12-0.11.5.pom
  download error: Caught java.io.IOException (Server returned HTTP response code: 502 for URL: https://dl.bintray.com/scalacenter/releases/org/scalameta/metals_2.12/0.11.5/metals_2.12-0.11.5.pom) while downloading https:
//dl.bintray.com/scalacenter/releases/org/scalameta/metals_2.12/0.11.5/metals_2.12-0.11.5.pom
  not found: https://oss.sonatype.org/content/repositories/public/org/scalameta/metals_2.12/0.11.5/metals_2.12-0.11.5.pom
  not found: https://oss.sonatype.org/content/repositories/snapshots/org/scalameta/metals_2.12/0.11.5/metals_2.12-0.11.5.pom
```

yuilib:
#563 should fix the issue.
Can you try the latest version?

uchiiii:
I have a question probably related to this.

I got the error below when I tried to install Metal.
```
  1   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current                                                                                       
  2                                  Dload  Upload   Total   Spent    Left  Speed                                                                                         
  3   0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0                                                                                        
  4   0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0                                                                                        
  5 100 42577  100 42577    0     0  45513      0 --:--:-- --:--:-- --:--:-- 5933k                                                                                        
  6 Error: Could not find or load main class                                                                                                                              
[...]

PR #603, opened 2022-07-28
Fix metals installer

yuilib:
The first commit fixed the installer for metals (#566). 
The second one is a workaround for an issue which the latest metals 0.11.7 seems not compatible with vim-lsp as before ([prabirshrestha/vim-lsp/#1331](https://github.com/prabirshrestha/vim-lsp/issues/1331)).


mattn:
Thank you. Could you please fix some warnings.
https://github.com/mattn/vim-lsp-settings/pull/603/files

yuilib:
Because the warning looks like an exception case on https://github.com/koalaman/shellcheck/wiki/SC2086, I think it is ok as compromise to keep `${java_flags}` unquoted.
If it must be fixed, can change the shell used from sh to bash because sh doesn't have array?
Which way to fix (changing to bash or unquoted) should I choose? Or both?

mattn:
Thanks.

PR #604, opened 2022-07-31
Upgrade clangd install

brentyi:
Currently the clangd installer supports only up to version 11, but a lot's happened in llvm since then!

Main changes:
- More general/less hardcoded logic for checking for supported versions
- Added support for llvm versions 12 and 13

Tested only on Ubuntu 20.04.

### DIFF
--- a/installer/install-clangd.sh
+++ b/installer/install-clangd.sh
@@ -94,17 +94,23 @@ filename() {
   echo "clang+llvm-$version-$arch-$platform"
 }
 
+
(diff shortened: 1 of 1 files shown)


mattn:
Looks good. I'll look into it later.

mattn:
How about to add invalid version 0 at the last and check that all tries are failed?

brentyi:
Error message for the failure case makes sense, just added. :slightly_smiling_face:

mattn:
Thank you

PR #608, opened 2022-08-14
fix clojure-lsp installer: DL latest standalone binary

sakuraiyuta:
Found at: https://github.com/shun/ddc-vim-lsp/pull/16#issuecomment-1212962805

Currently `clojure-lsp` installer downloads `2021.02.01-20.37.52`.
It's too old to use new features, and users need to install JRE.
This makes more steps to use LSP.

Now we can use native standalone binary in latest version, so I create this PR.

### DIFF
--- a/installer/install-clojure-lsp.cmd
+++ b/installer/install-clojure-lsp.cmd
@@ -1,5 +1,22 @@
 @echo off
 
 setlocal
-set VERSION=2021.02.01-20.37.52
-curl -L -o clojure-lsp.cmd https://github.com/clojure-lsp/clojure-lsp/releases/download/%VERSION%/clojure-lsp
+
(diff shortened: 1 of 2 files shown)


mattn:
Thank you. Could you please fix this too?

https://github.com/mattn/vim-lsp-settings/blob/2934abae01563b710e89df7981a24aec946b01d9/installer/install-clojure-lsp.cmd#L4

sakuraiyuta:
@mattn Thanks for reply!

I fixed installer for Windows.

vokoscreen-2022-08-15_17-06-01

Please check code and merge if no problems there.

mattn:
LGTM

mattn:
Thank you

PR #611, opened 2022-08-22
Removed workaround for metals

yuilib:
https://github.com/prabirshrestha/vim-lsp/pull/1350 solved the issue that metals (>= 0.11.7) couldn't run, so that the workaround in metals installer is removed and now latest metals works correctly.


mattn:
Thank you

PR #614, opened 2022-09-04
Added Hyuga installer

sakuraiyuta:
This PR adds Hy Language Server for Hy 0.24.0.

https://github.com/sakuraiyuta/hyuga

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,6 +111,7 @@ You can change the directory to install servers by set `g:lsp_settings_servers_d
 | HTML              | html-languageserver                 |    Yes    |      Yes      |
 | HTML              | angular-language-server             |    Yes    |      Yes      |
 | HTML              | tailwindcss-intellisense            |    Yes    |      Yes      |
+| Hy                | hyuga                               |    Yes    |      Yes      |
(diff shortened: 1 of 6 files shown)


mattn:
Thanks your contribution. Why don't you use pip_install.cmd ?

sakuraiyuta:
@mattn Thank you for reply.

> Could you please explain why `--system-site-package` is required?

`Hyuga` supports to search system installed packages. This feature needs to use `--system-site-package`.
In my opinion, some python developer want to use them.
(Yes, it's so dirty...but a way for users can dare to select.)

It seems no methods to path through option arguments to `pip_install.{sh,cmd}`, so I avoided to use `pip_install` script.
Any ideas? Or should drop it?

Other way: Add options to lsp settings in `initialize`, e.g. `additional-path: {system installed packages dir}` like `root-uri`.

mattn:
Thanks. So could you please update README `Local Install` to No? (If my understainding is right)

mattn:
Now ready to merge this?

sakuraiyuta:
Sorry for waiting you...ready now!

mattn:
Thank you!!

PR #616, opened 2022-09-10
Update clangd to 15

sago35:
I would like to use Clangd semantic highlighting on Windows.
The original code cannot be used on Windows because clangd 10.0.0 is installed.
This PR updates Windows installation scripts.
Scripts for linux and macOS have also been updated to use clangd 15.0.0 if the file is available.

https://github.com/llvm/llvm-project/releases/

### DIFF
--- a/installer/install-clangd.cmd
+++ b/installer/install-clangd.cmd
@@ -1,7 +1,7 @@
 @echo off
 
 setlocal
-set VERSION=10.0.0
+set VERSION=15.0.0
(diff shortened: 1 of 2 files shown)


sago35:
I tried on ubuntu20.04 and it installed clangd 13.0.0, the same as before the change.
Since LLVM 14.0.0, no files have been released for ubuntu 20.04, so it works as expected.

mattn:
Thank you

PR #619, opened 2022-09-14
Update volar to v0.40.13

tsukkee:
Also update its configuration.

### DIFF
--- a/installer/install-volar-server.cmd
+++ b/installer/install-volar-server.cmd
@@ -1,5 +1,5 @@
 @echo off
 
-call "%~dp0\npm_install.cmd" vue-language-server @volar/vue-language-server@0.40.1
+call "%~dp0\npm_install.cmd" vue-language-server @volar/vue-language-server@0.40.13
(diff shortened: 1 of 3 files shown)


mattn:
LGTM, BTW What is documentColor?

tsukkee:
Umm, sorry but to be honest I don't understand this option.
I just align it to the default values in Volar.
https://github.com/johnsoncodehk/volar/blob/d6c433ca479acca7b5c30d82f0e247fb8f69d6a4/examples/svelte/src/client.ts#L40

This option is described in [the LSP specification](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#documentColorClientCapabilities) but I assume that there is no special meaning in Volar because the parameter is just passed to the capability object but it is not referred in Volar code...
https://github.com/johnsoncodehk/volar/blob/a883a4763d709aff86d9c4aac2c9cbf7ed1d04dd/packages/language-server/src/registers/registerDocumentFeatures.ts#L17-L19

mattn:
Okay, Thanks.

PR #620, opened 2022-09-16
Add nix support

klarkc:


mattn:
Thanks your contribution. What features do the the language server provide? (for my interesting)

klarkc:
> Thanks your contribution. What features do the the language server provide? (for my interesting)

I am not an active user of `rnix-lsp`, I just started using it. But, for me, code completion and formatting are really helpful to write nix code. There are no docs tough, probably because the language lack of types.

settings.json
```diff
@@ -849,6 +849,16 @@
       }
     }
   ],
+  "nix": [
+    {
+      "command": "rnix-lsp",
+      "url": "https://github.com/nix-community/rnix-lsp",
+      "description": "Language server implementation for Nix",
+      "requires": [
+        "nix"
```
mattn, line 858: nix-env is also required?

klarkc: Yes, but as far as I know nix-env command is included into nix, so having nix means that you also have nix-env

mattn:
Thank you

PR #623, opened 2022-09-24
Fix purescript backed by purs-nix

klarkc:
Sometimes purescript runs on nix (with [purs-nix](https://github.com/ursi/purs-nix) for example), and in this case `spago.dhall` is not useful to detect `root_uri`, so we use `flake.nix` instead.


mattn:
Thank you

PR #627, opened 2022-10-17
Update Volar to v1.0.0

tsukkee:
Update Volar to v1.0.0 (actually v1.0.X). Since Volar is now stable, its version is specified by `~1.0.0`.

Also, `initializationOptions` becomes very simple because Volar now recognizes client capabilities.
But, now some hacks are needed because of https://github.com/prabirshrestha/vim-lsp/pull/1379 .

In addition, I deleted experimental multiple servers feature because Volar's architecture is changed.
I will try it again later...


mattn:
Thank you

PR #628, opened 2022-10-19
Omnisharp-LSP: Fix macOS .NET6 install

chaos95:
Currently, installing omnisharp-lsp on a macOS machine with .NET 6.0 fails due to a change in the naming scheme for the project's downloads. Starting with their .NET 6.0 builds, the osx builds are now provided for both arm64 and x64, which requires an additional specifier in the download filename. This change adds that specifier when installing the .NET 6.0 version of the LSP on a macOS machine.


mattn:
Thank you

PR #634, opened 2022-10-26
Fix Windows installation when username has space in it

chaos95:
This change modifies run_unzip.cmd and run_gzip.cmd to wrap invocations and paths in quotes.

This prevents issues arising where a user's home directory has spaces in it due to their username, such as 'C:\Users\FirstName LastName', which causes commands to fail because they parse the different parts of the path as separate arguments.

This change is not a complete solution - there are still quite a few individual installer scripts that make use of `%~dp0` without quotes, but these central changes to unzip and gzip should solve the problem for at least a few people. I would be happy to submit future PRs for the individual installers as time permits.

### DIFF
--- a/installer/run_gzip.cmd
+++ b/installer/run_gzip.cmd
@@ -2,8 +2,8 @@
 
 where gzip 2>NUL
 if %ERRORLEVEL% equ 0 (
-    gzip -d %*
+    gzip -d "%*"
 ) else (
-    curl -L -o %~dp0\busybox.exe https://github.com/mattn/vim-lsp-settings/releases/download/v0.0.1/busybox.exe
-    %~dp0\busybox gzip -d %*
+    curl -L -o "%~dp0\busybox.exe" https://github.com/mattn/vim-lsp-settings/releases/download/v0.0.1/busybox.exe
+    "%~dp0\busybox" gzip -d "%*"
 )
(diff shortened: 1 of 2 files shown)


mattn:
I suspect that this modification will make it even worse.

Current implementation:

test.c
```
#include <stdio.h>

int
main(int argc, char* argv[]) {
  int i = 0;
  for (i = 0; i < argc; i++) {
    puts(argv[i]);
  }
  return 0;
}
```
test.bat
```
@echo off
.\test.exe %*
```

```
C:\test>test.bat "foo bar"
foo bar
```

And with change

test.bat
```
@echo off
.\test.exe "%*"
```
```
C:\test>test.bat "foo bar"
foo
bar
```

chaos95:
True. I think it might be sufficient to wrap the commands in quotes and leave the arguments unquoted. For all of the invocations of `unzip.cmd` that I can see, the arguments are relative paths, so there are no spaces in them. But because the path to `unzip.cmd` itself might contain spaces, that still needs to be quoted. I'll update the PR shortly :)

mattn:
Sorry for the long silence, and thanks for the fix. Unfortunately quoting %* as a single string breaks callers that pass multiple arguments — e.g. installer/install-qmlls.cmd calls run_unzip.cmd with a zip name and a member to extract, which would become one quoted argument. Closing as-is; a version that quotes each argument separately is welcome.

mattn:
Sorry for the long silence, and thanks for the fix. Unfortunately quoting %* as a single string breaks callers that pass multiple arguments — e.g. installer/install-qmlls.cmd calls run_unzip.cmd with a zip name and a member to extract, which would become one quoted argument. Closing as-is; a version that quotes each argument separately is welcome.

PR #635, opened 2022-10-30
Update texlab.vim

tani:
The prefix of the workspace configuration for texlab is not 'latex' and is 'texlab'.
Cheers,


mattn:
Thank you

PR #636, opened 2022-11-04
Add typescript.tsx to settings.json

amadeus:
I was wondering why typescript.tsx file format was not working, and eventually realized it needed to be reflected in settings.json. 

Basically I've duplicated what is there for `typescriptreact`

### DIFF
--- a/settings.json
+++ b/settings.json
@@ -1575,6 +1575,46 @@
       "requires": []
     }
   ],
+  "typescript.tsx": [
+    {
+      "command": "typescript-language-server",
+      "url": "https://github.com/typescript-language-server/typescript-language-server",
+      "description": "TypeScript & JavaScript Language Server",
+      "requires": [
+        "npm"
+      ],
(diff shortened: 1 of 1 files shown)


amadeus:
Bump on this? 😬

settings.json
```diff
@@ -1575,6 +1575,46 @@
       "requires": []
     }
   ],
+  "typescript.tsx": [
```
mattn, line 1578: As far as I understand, This name does not work.

amadeus: I've been using this convention for years with my own typescript plugin and it worked fine?

mattn:
Sorry for leaving this open for so long. The server list for typescript/typescriptreact has changed a lot since this was opened (vtsls, biome, oxlint, etc.), so duplicating the entries for each compound filetype is hard to keep in sync. I'd rather handle compound filetypes like typescript.tsx generically in the loader. Closing for now — a fresh PR is welcome.

PR #637, opened 2022-11-04
README: add marksman information

tsuyoshicho:
Fix README.

Only add marksman.
I haven't checked the other "ls" (not listed, overlooked).


mattn:
Thank you

PR #638, opened 2022-11-05
Fix build error of ocaml-lsp

ushitora-anqou:
LspInstallServer for OCaml currently fails because ocaml-lsp requires OCaml v4.14.0. This patch fixes this issue.


mattn:
Thank you

PR #639, opened 2022-11-05
docs(readme): astro is added to the list

wuelnerdotexe:
@mattn - last time I didn't add `astro-ls` to the list of supported language servers (sorry), could you please accept it? :)


mattn:
Thanks

Issue #642, opened 2022-11-08
Format for taplo release assets has changed


dotdash:
As of the 0.8 release, the taplo lsp server can't be installed using `:LspInstallServer` anymore, because the release assets are no longer `.tar.gz` files, but just `.gz` files. The format of the file names have also changed and more platforms are supported.

mattn:
The gzipped file seems to be broken?

![image](https://user-images.githubusercontent.com/10111/200613162-5d66da31-4676-4896-beb2-c0b0074b295d.png)

thingsiplay:
Came here to report the same thing. Doing `:LspInstallServer` on a .toml file gets me this output: "!/home/tuncay/.vim/plugged/vim-lsp-settings/installer/install-taplo-lsp.sh"

```
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100     9  100     9    0     0     22      0 --:--:-- --:--:-- --:--:--    22

gzip: stdin: not in gzip format
tar: Child returned status 1
tar: Error is not recoverable: exiting now
```

I am on Manjaro and my stuff is usually updated (including Vim and all plugins). Lsp Servers for other formats work fine work fine, so it is only this particular taplo server I can't install.

buihachi-nishi:
I have checked the latest installer. For m1/m2 Macs, the output of uname -m is 'arm64', which results in an incorrect download link being generated (with the filename 'taplo-full-darwin-aarch64.gz')

nami-minami:
taplo installer is broken, and my environment is Darwin X86_64.
<img width="644" alt="スクリーンショット 2024-03-01 16 26 59" src="https://github.com/mattn/vim-lsp-settings/assets/111250887/d333266c-b0ad-4194-ac93-87e967ccaa4a">
The reason is that taplo 13.0 release does not provide binary. You can find related information at [](https://github.com/tamasfe/taplo/releases)
It seems that we need to build the source code.

frank770504:
I also found an issue with toml.

Currently, in the latest release, it drops "full" from the file name, so the curl script cannot grab the binary.

- For example
  - The previous name: [taplo-full-linux-x86_64.gz](https://github.com/tamasfe/taplo/releases/download/0.9.3/taplo-full-linux-x86_64.gz)
[...]

PR #649, opened 2022-12-28
Fixed installer to work with latest taplo release (0.8.0)

apelloni:
This fixes the problem with the new format of the release files as discussed in #642
I didn't change the `install-taplo-lsp.cmd` installer because I have no way to test it


mattn:
Thank you

PR #650, opened 2023-01-02
Add a mention about fsautocomplete in README.md

KiYugadgeter:
There is no mention about fsautocomplete despite it is already supported.


mattn:
Good catch. Thank you

PR #651, opened 2023-01-02
Remove --background-service-enabled and add --adaptive-lsp-server-enabled

KiYugadgeter:
This is based on https://github.com/neovim/nvim-lspconfig/pull/2197


mattn:
Thank you

PR #652, opened 2023-01-21
improve installers to pass shellcheck checks

itchyny:
The linter `shellcheck` checks are ignored in some merged PRs but I think we can fix most of the warnings. I removed the `.shellcheckrc` file to run `shellcheck` more strictly and fixed all of the warnings. I also run `make shfmt` to reformat all the installers.


mattn:
Thanks!

PR #656, opened 2023-02-21
New url for terraform-ls downloads

z0rc:
terraform-ls doesn't publish binaries under github releases anymore. Update url to reference releases.hashicorp.com.

I don't have environment to test Windows changes, hopefully someone will be able to confirm it's working.


mattn:
Thank you

PR #660, opened 2023-03-09
Escape the buffer path for whitespace in path

lkebin:
When my project path includes whitespace (/home/kb/Hello World/), the blocklist works unexpectedly.


mattn:
I think `lsp#utils#find_nearest_parent_file_directory` should be fixed.

lkebin:
> I think `lsp#utils#find_nearest_parent_file_directory` should be fixed.

I have created a PR to fix this function from vim-lsp repo (prabirshrestha/vim-lsp/#1444). I'm not sure whether it caused the downstream plugin error. 

If the downstream plugin using `fnameescape`, will have twice escape operation, may not work as expected.

mattn:
Closed by https://github.com/prabirshrestha/vim-lsp/pull/1444

PR #667, opened 2023-03-31
Fix 'taplo-full-darwin-arm64.gz' to 'taplo-full-darwin-aarch64.gz'

buihachi-nishi:
I have corrected the filename from 'taplo-full-darwin-arm64.gz' to 'taplo-full-darwin-aarch64.gz'.


buihachi-nishi:
closes #642

mattn:
Thank you

PR #668, opened 2023-04-01
Allow volar-server to handle typescript files

flo-at:
volar-server offers to handle ts files which makes *.vue imports work without the TypeScript Vue Plugin.
See: https://vuejs.org/guide/typescript/overview.html#volar-takeover-mode


mattn:
Thank you

PR #674, opened 2023-04-28
Updated ts in volar installer to solve auto import issue

DanielViberg:
Solves a bug where auto import works poorly in volar-server. 
See https://github.com/vuejs/language-tools/issues/2190 for more info.


mattn (approved):

mattn:
Thank you

PR #677, opened 2023-05-16
Update elixir-ls version to 0.14.6

pojiro:
ref. [Release v0.14.6](https://github.com/elixir-lsp/elixir-ls/releases/tag/v0.14.6)


mattn:
Thank you

PR #678, opened 2023-06-04
fix installation of marksman on linux

stefanSchinkel:
The download URL for marksman has changed as of https://github.com/artempyanykh/marksman/releases/tag/2023-06-01 and now for Linux the arch needs to be specified as well. Updated the marksman install script to switch between x86_64 and arm64. 
*NOTE*: I only had a x86 box readily available on which I tested the change. Didn't test for ARM, but since changes are minimal it should work.


mattn:
Thank you

PR #681, opened 2023-06-07
Change godot port to 6005

idbrii:
Godot changed the LSP port for Godot 4:
https://github.com/godotengine/godot/blob/master/modules/gdscript/language_server/gdscript_language_server.h#L47-L48

The change was in this commit:

https://github.com/godotengine/godot/commit/de7873c2d87c3ee8af8c27e35f49767fcc384e75

Godot 3 users can change to use this port in editor -> editor settings -> network -> language server -> remote port. Or they can use `let g:lsp_settings.godot = { 'tcp': '127.0.0.1:6008'}` to use the old port instead.

### DIFF
--- a/settings/godot.vim
+++ b/settings/godot.vim
@@ -2,7 +2,7 @@ augroup vim_lsp_settings_godot
   au!
   LspRegisterServer {
       \ 'name': 'godot',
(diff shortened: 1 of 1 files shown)


mattn:
Thank you.

PR #683, opened 2023-06-20
Add nixd alternative

klarkc:
[nixd](https://github.com/nix-community/nixd) is a brand new LSP alternative for nix supporting official nix libs.


mattn:
Thanks. Could you please add `install-nixd-lsp.cmd` ?

```dosbatch
@echo off

call nix-env -i -f "https://github.com/nix-community/nixd/archive/master.tar.gz"
```

mattn:
Thank you.

PR #684, opened 2023-06-22
Update volar to v1.8.0

tsukkee:
Its npm package was moved to https://www.npmjs.com/package/@vue/language-server


mattn:
Good catch. Thanks.

PR #685, opened 2023-06-27
Add default efm config

kflu:
efm-langserver is no use without a config.yaml. This PR adds a default config.yaml if one does not exist.


mattn:
No use default config file is intended. But I'll wait anyone's opinions.

PR #687, opened 2023-07-28
Update elixir-ls version to 0.15.1

pojiro:
ref. [Release v0.15.1](https://github.com/elixir-lsp/elixir-ls/releases/tag/v0.15.1)


mattn:
Thank you

PR #689, opened 2023-07-31
Use `uname -m` as `-p` doesn't always work

jm33-m0:
See https://unix.stackexchange.com/questions/307955/uname-p-i-are-unknown `uname -m` returns machine type, which is enough for this script, `uname -p` doesn't work on most Linux distros.

In my system (Arch), executing `uname -p` returns `unknown` therefore the script downloads ARM64 version of `marksman`.


mattn:
Thank you

PR #690, opened 2023-08-02
Fix elixir-ls installer

pojiro:
From v0.15.0 [elixir-ls](https://github.com/elixir-lsp/elixir-ls) doesn't provide `elixir-ls.zip`.
Now provided by `elixir-ls-[VERSION].zip`.


simlay:
@mattn up to merge this? I was about to author an almost identical PR myself.

mattn:
Sorry my delay. Thank you

PR #691, opened 2023-08-02
fix: Correct elixir-ls release urls

jHwls:
Fixes the elixir-ls installer failing :)


mattn:
fixed by https://github.com/mattn/vim-lsp-settings/commit/9f8fc440b72f50d9d1d250d4ad1f8ac4e14cc152

thanks.

PR #693, opened 2023-08-04
gopls: change capabilities to use documentSymbol

satorunooshie:
instead of deprecated symbolInformation.

https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#symbolInformation

```
/**
 * Represents information about programming constructs like variables, classes,
 * interfaces etc.
 *
 * @deprecated use DocumentSymbol or WorkspaceSymbol instead.
 */
```


mattn:
Thank you

PR #700, opened 2023-09-17
tailwindcss-intelisense version 0.5.10 -> 0.10.0

gawetto:
Auto-complete did not work with tailwindcss-intelisense, but it worked after using the latest version (v0.10.0)


mattn:
Thank you

PR #705, opened 2023-10-23
new file format for ntt archive for windows

moosq:
From `ntt v0.18.0` on the archive deployed for windows operating system will be in `. zip` format


mattn:
Thank you

PR #706, opened 2023-11-15
Fix deno's inlay hint settings

heavenshell:
deno's inlay hint did not work since deno v1.38.0.

before:
スクリーンショット 2023-11-16 1 04 33

after:
スクリーンショット 2023-11-16 1 03 40

Thanks to @yaegassy for suggested me to how to fix this problem.


mattn:
Thank you

Issue #708, opened 2023-11-23
install-terraform-ls.sh curling non-existant file


drichardson:
`LspInstallServer` failed to install terraform-ls. When I ran the `install-terraform-ls.sh` script manually after setting `set -x` to show commands being run, I saw this:

```bash
[I]$ ./install-terraform-ls.sh
++ uname -s
++ tr '[:upper:]' '[:lower:]'
+ os=darwin
++ uname -m
++ tr '[:upper:]' '[:lower:]'
+ arch=arm64
+ case $os in
+ case $arch in
+ arch=arm64
+++ curl -Ls -o /dev/null -w '%{url_effective}' https://github.com/hashicorp/terraform-ls/releases/latest
++ basename https://github.com/hashicorp/terraform-ls/releases/tag/v0.32.3
+ version=v0.32.3
++ echo v0.32.3
++ cut -c2-
+ short_version=0.32.3
+ filename=terraform-ls_0.32.3
+ url=https://github.com/hashicorp/terraform-ls/releases/download/v0.32.3/terraform-ls_0.32.3_darwin_arm64.zip
+ filename=terraform-ls_0.32.3.zip
+ curl -L --progress-bar https://github.com/hashicorp/terraform-ls/releases/download/v0.32.3/terraform-ls_0.32.3_darwin_arm64.zip -o terraform-ls_0.32.3.zip
################################################################################################ 100.0%
+ unzip terraform-ls_0.32.3.zip
Archive:  terraform-ls_0.32.3.zip
  End-of-central-directory signature not found.  Either this file is not
  a zipfile, or it constitutes one disk of a multi-part archive.  In the
  latter case the central directory and zipfile comment will be found on
  the last disk(s) of this archive.
unzip:  cannot find zipfile directory in one of terraform-ls_0.32.3.zip or
        terraform-ls_0.32.3.zip.zip, and cannot find terraform-ls_0.32.3.zip.ZIP, period.
doug@dougs-mbp aws=eng_user k8s=dev+main-backend ~/.vim/pack/common/start/vim-lsp-settings/installer (9)
[I]$ ls *.zip
terraform-ls_0.32.3.zip
doug@dougs-mbp aws=eng_user k8s=dev+main-backend ~/.vim/pack/common/start/vim-lsp-settings/installer
[I]$ cat terraform-ls_0.32.3.zip
[...]

PR #709, opened 2023-11-23
Fix install-terraform-ls.sh

drichardson:
Hashicorp is releasing binaries on releases.hashicorp.com instead of github releases. Update install-terraform-ls.sh to pull from the new location.

Fixes https://github.com/mattn/vim-lsp-settings/issues/708

## Testing

Before changes, `install-terraform-ls.sh` failed. The downloaded zip file was actually an ASCII file (not a zip) consisting of `Not found`.

After changes, confirmed working with:

```bash
$ ./install-terraform-ls.sh
################################################################################################ 100.0%
Archive:  terraform-ls_0.32.3.zip
replace terraform-ls? [y]es, [n]o, [A]ll, [N]one, [r]ename: y
  inflating: terraform-ls
$ echo $?
0
```

### DIFF
--- a/installer/install-terraform-ls.sh
+++ b/installer/install-terraform-ls.sh
@@ -24,10 +24,9 @@ arm64*) arch=arm64 ;;
   ;;
 esac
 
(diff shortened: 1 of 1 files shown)


drichardson:
nevermind, already fixed.

PR #710, opened 2023-12-08
Add Ruby Sorbet

tsukkee:
I add [Sorbet](https://sorbet.org/) settings for vim-lsp.

I adopt similar way to RuboCop (LSP mode) setting to create sorbet's one because sorbet executable (`srb` command) is expected to be installed by Bundle in the workspace according to the reference ( https://sorbet.org/docs/vscode#installing-and-enabling-the-sorbet-extension ).

Also, Sorbet needs watchman to watch file changes and if that is not installed `srb` command needs to have `--disable-watchman` flag.

NOTE: CodeAction is NOT worked correctly maybe due to Sorbet's issue or vim-lsp's issue. I'm trying to solve this now.


mattn:
Thank you

PR #711, opened 2023-12-15
Fix deno lsp did not work when using v1.39.0

kuuote:
All `initialization_options` are always overwritten `workspace_config` at deno lsp.
So moving `initialization_options` to `workspace_config.deno`.

It's works with deno v1.29.0(released over 1 years ago).


mattn:
Thank you

PR #712, opened 2023-12-20
Remove parameters for vala language server

### LINKED ISSUE #675: Incorrectly adding --stdio to command
Reference:
https://github.com/prabirshrestha/vim-lsp/issues/1471

This line is wrong:
https://github.com/mattn/vim-lsp-settings/blob/f98467680e524109767c795547e115d6e35425a9/settings/vala-language-server.vim#LL5C5-L5C191

It should not have the `'--stdio'` argument.

Thank you for your efforts!

albfan:
Seems there was a massive addition of this parameter `--stdio`.

I'm unsure if that is something ls should support but in vala-language-server is not supported (I can report if that should be supported)

fixes #675

### DIFF
--- a/settings.json
+++ b/settings.json
@@ -1671,11 +1671,7 @@
       "url": "https://github.com/Prince781/vala-language-server",
       "description": "Code Intelligence for Vala & Genie",
       "requires": [
-        "meson",
         "valac"
-      ],
-      "root_uri_patterns": [
-        "meson.build"
       ]
     }
   ],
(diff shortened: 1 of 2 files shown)


albfan:
Any non trivial project using vala will have meson.build in each subdirectory, so search for meson.build is wrong.

require meson to build is pretty command but not general. Is better to remove it

mattn:
Thank you

PR #713, opened 2023-12-23
Fix: Python version check for pip (in Python 3.12 or later, removed distutils)

tsuyoshicho:
## Background

Python 3.12 or later, distutils has been removed.
See: https://docs.python.org/3/whatsnew/3.12.html

So the previous version check code will fail.

## Changes

Change this to a simple check and fix it.
(Currently, the use of Python 2 is declining, so I thought this was a good idea.)


mattn:
Thank you

PR #717, opened 2024-01-05
ci: update checkout actions to a non-deprecated node version

storopoli:
We are getting tons of warnings in CI:

> The following actions uses node12 which is deprecated and will be forced to run on node16: actions/checkout@v2. For more info: https://github.blog/changelog/2023-06-13-github-actions-all-actions-will-run-on-node16-instead-of-node12-by-default

Example: https://github.com/mattn/vim-lsp-settings/actions/runs/7344860647


mattn:
Thank you

PR #728, opened 2024-03-12
Update Volar (currently called Vue Language Tools) to v2.0

tsukkee:
Vue Language Tools v2.0 consists of the `@vue/language-server` and `@vue/typescript-plugin`.

* Add `g:lsp_settings_filetype_vue` setup instruction to enable both `volar-server` and `typescript-language-server` in `vue` filetype.
* Add `@vue/typescript-plugin` to `typescript-language-server` plugins only in Vue projects.
* Show warnings when the setup is incomplete.

I think this is the first case to utilize typescript language server's plugins in vue-lsp-setttings.
So, I would like to have advice if my approach to do that is not appropriate.

### DIFF
--- a/settings/typescript-language-server.vim
+++ b/settings/typescript-language-server.vim
@@ -1,3 +1,38 @@
+function! s:find_vue_plugin() abort
+  let package_json_path = lsp#utils#find_nearest_parent_file(lsp#utils#get_buffer_path(), 'package.json')
+  if empty(package_json_path)
+    return v:null
+  endif
+
+  let package_json = json_decode(join(readfile(package_json_path), ''))
+  if !(has_key(package_json, 'dependencies') && has_key(package_json['dependencies'], 'vue'))
+    return v:null
+  endif
+
+  let plugin_location = lsp_settings#servers_dir() .. '/volar-server/node_modules/@vue/typescript-plugin'
+  if !isdirectory(plugin_location)
+    call lsp_settings#utils#warning('Please install the latest volar-server to enable Vue support')
+    return v:null
+  endif
+
+  return {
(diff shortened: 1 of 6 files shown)


settings/typescript-language-server.vim
```diff
@@ -1,3 +1,38 @@
+function! s:find_vue_plugin() abort
+  let package_json_path = lsp#utils#find_nearest_parent_file(lsp#utils#get_buffer_path(), 'package.json')
```
heavenshell, line 2: [ask]
What happens if there is vue in the root package.json In monorepo?
It would be nice if users could inject the location of package.json(global variable or can get from `lsp_settings#get()?`

tsukkee: Thank you for your feedback! I couldn't consider that… 
What I want to do is enabling `@vue/typescript-plugin` only in Vue projects, but I don't come up with an idea to detect Vue projects even in monorepo situation 🤔 
Now I'm wondering if `@vue/typescript-plugin` is always enabled when it exists, since it doesn't have bad side effects (might have small performance degradation, though).

tsukkee: I modified to use `@vue/typescript-plugin` when it just exists https://github.com/mattn/vim-lsp-settings/pull/728/commits/983dc1666b645451312f026307d345e79a4f9b97 .

I think that this still has small side effect to all users since only Vue users will install `volar-server`.

mattn:
@heavenshell any opinion?

heavenshell:
@mattn I'm sorry did not notice your mention.
LGTM!

mattn:
Thank you

PR #730, opened 2024-03-17
Find instelled clangd from PATH

hidai:
Most of the clangd binaries that can be downloaded from GitHub releases are not available for amd64, and it is recommended to use the ones provided by the distribution. Currently, the latest amd64 binary installed by vim-lsp-settings is 13, but Ubuntu 22.04 has 15.

This PullRequest uses the binaries provided by the distribution, if any.


mattn:
Thank you

PR #738, opened 2024-04-26
add biome

Milly:
Following as a reference for implementation.
https://biomejs.dev/guides/integrate-in-editor/
https://github.com/neovim/nvim-lspconfig/blob/master/doc/server_configurations.md#biome


mattn:
Thank you

PR #753, opened 2024-06-03
fixbug: digestif init argument error

p1ain0:
In Mac, digestif error with:"Failed to initialize digestif with error 1: /usr/local/texlive/2023/texmf-dist/scripts/digestif/digesti:110: Invalid configuration option: diagnostics = "true"


mattn:
Can you change like this?

```
has('macunix') ? v:null : {'diagnostics': 'true'}
```

p1ain0:
Okay, I have changed. Thank you!

mattn:
Thank you

ylee88:
I encountered the same error but on the Arch Linux. Apparently `{'diagnostics': 'true'}` is not a valid option for `digestif`?
```sh
❯ digestif --version
Digestif 0.5.1
```

PR #761, opened 2024-07-12
Support expandMacro for rust-analyzer lsp

simlay:
There are a bunch of LSP extensions rust-analyzer has. This project implements a few of them. One that it doesn't is [`rust-analyzer/expandMacro`](https://github.com/rust-lang/rust-analyzer/blob/master/docs/dev/lsp-extensions.md#expand-macro).

Full disclosure, I'm a novice at vimL. I copied the `on_rust_analyzer_status` function and modified it for the response handler. I won't feel bad about a strong constructive critique.

Below is an example of what my vim looks like when I expand the `println!` macro. I've got a powerbar plugin at the bottom. The split for the expanded macro yields source that frequently has more macros. Unfortunately, there's no file for that source so I'm not sure it's macro-expandable? I'd love to be able to expand each macro down the chain but my vimscript/vimL is weak.

image


settings/rust-analyzer.vim
```diff
@@ -159,6 +162,47 @@ function! s:on_find_matching_brace(x) abort
...
+    let l:alternate=@#
+    silent! pclose
+    sp LspRustExpandMacro
+    execute 'resize '.min([len(l:lines), &previewheight])
+    set previewwindow
+    setlocal conceallevel=2
+    setlocal bufhidden=hide
+    setlocal nobuflisted
+    setlocal buftype=nofile
+    setlocal ft=rust
```
heavenshell, line 197: Hi, Thank you for great work!
I've try this ane love it. But I have some feed backs.

I use `g:lsp_diagnostics_virtual_text_enabled = true`.
Execute `:LspRustExpandMacro` and move to preview window.
Shows a lot of errors and bit a annoying 😰 

https://github.com/user-attachments/assets/efd87f7b-5cb3-4b1e-8b27-ccb6e7847cc4

I'm not sure my suggestion is right or not, but set no filetype and enable syntax for preview looks good to me.

```diff
diff --git a/settings/rust-analyzer.vim b/settings/rust-analyzer.vim
index b71d6eb..dde8fd7 100644
--- a/settings/rust-analyzer.vim
+++ b/settings/rust-analyzer.vim
@@ -194,10 +194,13 @@ function! s:on_expand_macro(x) abort
     setlocal bufhidden=hide
     setlocal nobuflisted
     setlocal buftype=nofile
-    setlocal ft=rust
     setlocal noswapfile
     %d
     call setline(1, l:lines)
+
+    let l:win_id = win_getid()
+    call win_execute(l:win_id, $'setlocal syntax=rust')
+
     execute "normal \<c-w>p"
     call winrestview(l:view)
     let @#=l:alternate
```

https://github.com/user-attachments/assets/fa25d652-0d89-4430-b955-0fc4e4690c5f

@mattn Do you have any good solutions?

simlay: Oh nice! I don't have the virtual diagnostics enabled but definitely see how it's rough. I think the syntax set to rust works. I've added those changes.

simlay:
I was curios why the expansions always yield `$crate` but that crate is different depending on where the macro came from. `$crate` from `println!` is `std`, but with `log::info!`/`trace::info!`, it's `log` or `trace`. In any case, I don't think the client is suppose to swap those. https://github.com/rust-lang/rust-analyzer/issues/7532 is the issue.

Poking around, some folks may want the macro to expand inline (https://github.com/rust-lang/rust-analyzer/issues/11888#issuecomment-1890985494). That may be a feature for the future as it'd help with being able to manually recurse through the expansions. Given that the macro expansions have `$crate` in them, I think putting them in a separate/preview split is best for now.

simlay:
I was about to author a Rust Analyzer LSP Extensions tracking issue (really just a matrix of extensions) when I realized it should probably just go in the notes section of the README along with [an example of a `.vim-lsp-settings/settings.json`](https://github.com/simlay/ferris-ui/blob/d6f203a2f044b4da3dd6240040181baec5152ebd/.vim-lsp-settings/settings.json). Thoughts?

simlay:
@mattn Ping. Any requests on this PR? I find the feature pretty useful.

mattn:
Thank you

PR #762, opened 2024-07-12
Added rust-analyzer notes to README

simlay:
In #761, I figured it'd be nice to write down the status of the various rust-analyzer extension support.


mattn:
Thank you

PR #764, opened 2024-07-14
add -e switch at install-zls.sh

conao3:
I don't know why `set -e` is not set, but other installer have this switch and I think it is better to set.


conao3:
Check all files and add `set -e` if not added.

mattn:
Thank you

conao3:
Your welcome!

PR #765, opened 2024-07-14
elixir is not needed for installing elixir-ls

conao3:
`elixir` is not needed for installing `elixir-ls`
link:
- https://github.com/mattn/vim-lsp-settings/blob/master/installer/install-elixir-ls.sh
- https://github.com/mattn/vim-lsp-settings/blob/master/installer/install-elixir-ls.cmd


mattn:
thank you

PR #771, opened 2024-08-20
Add starpls server (Starlark LSP)

kmARC:
"starpls is a language server for Starlark, the configuration language used by Bazel and Buck2." [1]

[1]: https://github.com/withered-magic/starpls


kmARC:
Hi @mattn,

I'm using starpls for a while, and it works perfectly with this configuration. It would be a nice addition to the supported LSP servers. This PR is very low-effort by now (undocumented, untested); so please let me know if you are willing to consider the addition, and what else is required.

mattn:
Thank you

PR #789, opened 2025-01-13
fix: do not enable steep if Steepfile is not exist to resolve #756

### LINKED ISSUE #756: The steep language server started when Steepfile not exists
My Vim has these settings to enable the `ruby-lsp`, `steep`, and `rubocop` language server

```
let g:lsp_settings_filetype_ruby = ['ruby-lsp', 'rubocop', 'steep']
```

However, when I installed steep with `gem install steep` another project didn't have `Steepfile` still load the steep language server.

I think this line will check for it but seems not to work:
https://github.com/mattn/vim-lsp-settings/blob/2cad696eb76755ff00415d97ce4bd5c82d6e2fb3/settings/steep.vim#L8

If I removed the `g:lsp_settings_filetype_ruby` then the steep not work on the project has `Steepfile`

> P.S. The `rubocop` or `rubocop-lsp-mode` both do not run the rubocop language server for now

elct9620:
According to https://github.com/mattn/vim-lsp-settings/issues/756#issuecomment-2307006065 the `lsp_settings#root_path(['Steepfile'])` cannot detect `Steepfile` correctly.

I referenced the `typescript-language-server.vim` and changed it to `lsp#utils#find_nearest_parent_file(lsp#utils#get_buffer_path(), 'Steepfile')`, and the issue was resolved.


mattn:
Thank you

PR #790, opened 2025-01-13
Add support for the qml language server

lassfido:
Hi,

The Qt company recently released a standalone version of the qmlls language server bundled with the Qt Extension for VS Code extension. I like vim-lsp and want to add support for the qml language.


mattn:
Thank you

PR #791, opened 2025-01-14
fix: Change rust-analyzer binary download URL for Windows.

ansanloms:
In the latest [rust-analyzer releases](https://github.com/rust-lang/rust-analyzer/releases), gzip Windows binaries are no longer included. 
Therefore, I have updated the installer to download the zip archive instead.


mattn:
Thank you

PR #795, opened 2025-01-27
Fix install-powershell-languageserver.cmd caused pwsh server no start

ErrorPower2001:
I'm not very familiar with the syntax in this cmd file, but I think ``%PSES_BUNDLE_PATH: =` %`` is not a standard cmd syntax. Also, moving the double quotes (which are ^" here) to the end caused PSES to remain in the "starting" state.


mattn:
Thank you

ErrorPower2001:
for everyone: This code will still cause it to fail to run. If you are using pwsh core, please use pwsh instead of powershell. I have already tested it on pwsh 7.

PR #798, opened 2025-03-24
Update elixir-ls version to 0.27.2

pojiro:
ref. [Release v0.27.2](https://github.com/elixir-lsp/elixir-ls/releases/tag/v0.27.2)


mattn:
Thank you

PR #799, opened 2025-03-31
Switch to pwsh

ErrorPower2001:
according to [Supported PowerShell Versions](https://github.com/PowerShell/PowerShellEditorServices?tab=readme-ov-file#supported-powershell-versions) and my tests, the powershell-language-server only work in pwsh core 7+.

I also modified the code to make the cmd file generated by the installer in CRLF format.


mattn:
Anyone have objection to switch powershell to pwsh ?

PR #803, opened 2025-05-16
Add bacon-ls

Milly:
[`bacon-ls`](https://github.com/crisidev/bacon-ls) is a Language Server for Rust using [`bacon`](https://dystroy.org/bacon/) diagnostics.


mattn:
Thank you

PR #806, opened 2025-07-02
Update Volar (Vue Language Tools) to v3.0.0

tsukkee:
There are 2 breaking changes according to the upgrade guide [^1].
* TypeScript Language Server and Vue Language Server now communicate on LSP protocol, previously they did it on named pipes, so I added some code to do that by follwoing Neovim example [^2]
  * And, now only vtsls is supported becuase typescript-language-server doesn't support this communication method right now
* `typescript.tsdk` option has dropped so I just removed it.

[^1]: https://github.com/vuejs/language-tools/discussions/5456
[^2]: https://github.com/vuejs/language-tools/wiki/Neovim

### DIFF
--- a/README.md
+++ b/README.md
@@ -429,10 +429,10 @@ For more details, please see [Sorbet](https://sorbet.org/docs/vscode#installing-
 
 ### [volar (Vue Language Tools)](https://github.com/vuejs/language-tools)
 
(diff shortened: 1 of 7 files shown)


mattn:
Thank you

PR #807, opened 2025-07-04
Check vtsls is used for Vue Language Tools (volar-server)

tsukkee:
`vtsls` should be used also in .ts file for cast that .ts file refers .vue file.
So, I added code to check the setting, and update README.


mattn:
Thank you

PR #810, opened 2025-07-08
Update taplo lsp settings

oliverhr:
Modified only for Windows:

- update Windows installer to support the latest version
- updated installed to detect windows os architecture


mattn:
Thank you

PR #825, opened 2025-10-20
Update buf

bufdev:
This PR does two things:

- Updates `buf` post-v1.59.0, where LSP was promoted to general availability. The command was changed from `buf beta lsp` to `buf lsp serve`.
- Delete `bufls` entirely. This was a prototype from years ago that was never finished: https://github.com/bufbuild/buf-language-server the repository is archived, and no one should use this.


bufdev:
Gentle ping here!

PR #837, opened 2026-02-08
Fix kotlin-lsp not running on linux and update to the latest version

declval:
Fixed kotlin-lsp not running on linux due to vim-lsp-settings expecting the file to be named kotlin-lsp while it was actually named kotlin-lsp.sh by making a symbolic link. Also made kotlin-lsp.sh executable and updated kotlin-lsp to the latest version.

Since kotlin-lsp starting from version [v261.13587.0](https://github.com/Kotlin/kotlin-lsp/releases/tag/kotlin-lsp%2Fv261.13587.0) provides platform-specific builds I added some logic to download the appropriate version.

Lastly in the changelog they say "no JDK required by default, the language server bundles its own" so maybe Java requirement may be dropped but I'm not familiar with Java nor with vim-lsp-settings enough to assess if that change would be correct.

### DIFF
--- a/installer/install-kotlin-lsp.cmd
+++ b/installer/install-kotlin-lsp.cmd
@@ -1,8 +1,14 @@
 @echo off
 
 setlocal
-curl -L -o server.zip "https://download-cdn.jetbrains.com/kotlin-lsp/0.252.17811/kotlin-0.252.17811.zip"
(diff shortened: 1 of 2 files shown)


mattn:
LGTM

PR #845, opened 2026-03-28
Fix install oxfmt from npm registry

heavenshell:
Hi, sorry for again and again 🙇 

oxfmt now requires the npm-distributed package for LSP, 
so this PR changes the installation method from downloading the standalone binary to installing via npm.


mattn:
Thank you

PR #855, opened 2026-07-11
Fix pin typescript@^6.0.0 as latests doesnt ship tsserver

EppurSiMu0ve:
As TypeScript 7 no longer ships with the legacy lib/tsserver.js or bin/tsserver files, and because typescript-language-server installation script specify no version, npm blindly grabs the latest release (which is now 7.0+).

So, for now, to get typescript-language-server working and automatically configured by vim-lsp-setting it is needed to lock the dependency to the 6.x release.

Source: [zed editor repository](https://github.com/zed-industries/zed/issues/60618)

### DIFF
--- a/installer/install-typescript-language-server.cmd
+++ b/installer/install-typescript-language-server.cmd
@@ -1,4 +1,4 @@
 @echo off
 
-call "%~dp0\npm_install.cmd" tsserver typescript
+call "%~dp0\npm_install.cmd" tsserver typescript@6
 call "%~dp0\npm_install.cmd" typescript-language-server typescript-language-server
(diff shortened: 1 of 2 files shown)


mattn:
Thanks! Merging.

For the record on why the Windows side needed a different range syntax: cmd.exe treats `^` as an escape character, so `typescript@^6.0.0` in the batch installer gets mangled while parsing — the caret is swallowed and it collapses to `6.0.0` (an exact pin), and doubling/quoting the caret instead produces an `EINVALIDTAGNAME` (`^^6.0.0`). Using the caret-free range `typescript@6` (`>=6.0.0 <7.0.0`) is equivalent to `^6.0.0`, survives cmd parsing, and resolves to the latest 6.x. The `.sh` side was switched to `@6` as well just to keep both installers consistent.

EppurSiMu0ve:
Thanks for the merging and for the great work you've done with this plugin!

